PR #39, opened 2017-10-25
Cleaning up declare_estimator

nfultz:
* helper function for arg sanitization
* move model specific chunk into helper
* consolidate model/estimator branches
* return estimand label as argument
* replace evals with match.fun


graemeblair:
swag. thanks, neil.

PR #54, opened 2017-11-15
Make design templates work with shiny

nfultz:
This updates the design library vignettes to work with the [WIP DDinspector](https://github.com/DeclareDesign/shiny/blob/nfultz/builder_inspector/R/inspector.R) shiny applet and the [memoise caching mechanism](https://github.com/DeclareDesign/DDtools/blob/master/R/design_library.R) it uses.

To make this work, there are a couple key changes:

  * template should be stored in a variable name `$(VIGNETTE_TOPIC)_template` so `regression_discontinuity_template` and not `rd_template`
  * the design library generator loads the template straight out of the vignette-generated R script ("R/regression_discontinuity_template.R") in the DeclareDesign package via `source()`, so that file needs to be able to run start to finish - I added some guard rails to the vignette code to deal with missing intermediate files that they might generate and use - we do not need to save out `.RDS` copies of the templates, (it's already too confusing trying to match up code to RDS :( ). 
  * formal arguments on the template will be handed off to `expand.grid` and each combination will be pre-computed
  * The first in the vector of arguments will be the default in the drop down in DDinspector 
  * right now DDinspector only supports numeric arguments, we can fix this in a later version
   * the RDS files in vignette will be excluded from the package builds as per cran policy on large-ish files.

I've also included a stub for two-way factorial designs if we want to include that in the demo.


graemeblair (approved):

graemeblair:
terrific, thanks, this looks great.

PR #56, opened 2017-11-16
Fix bare declares

nfultz:
This fixes bare declare_designs (not in variables), such as:

```
declare_design(
	declare_population(N=100,noise=rnorm(N)),
	declare_potential_outcomes(Y_Z_0=noise, Y_Z_1=noise+1),
	declare_estimand(ATE=mean(Y_Z_1 - Y_Z_0), label="ATE"),
	declare_sampling(n=20),
	declare_assignment(m=10),
	reveal_outcomes(),
	declare_estimator(Y~Z, estimand="ATE")
)
```

It also handles character estimand labels to make the above example join estimators to estimands correctly.

### DIFF
--- a/R/declare_design.R
+++ b/R/declare_design.R
@@ -104,17 +104,21 @@ declare_design <- function(...,
   name_or_call <- sapply(causal_order_expr, class)
 
   ## wrap any call in wrap_step()
-  if (length(dots) > 1) {
-    for (i in 2:length(dots)) {
-      if (name_or_call[[i]] == "call") {
-        dots[[i]] <-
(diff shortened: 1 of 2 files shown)


graemeblair:
that's so great. @nfultz will the code above work if the label and estimand args are unquoted?

nfultz:
label  on declare_estimand is unchanged. On declare_estimator, if the
estimand arg is unquoted and not a function, the label will be set to NULL
if estimand is an object and otherwise throw a 'not found' error, no NSE is
performed.

PR #58, opened 2017-11-20
fix copy paste error, add main effects to interaction in 2 way factorial

nfultz:


graemeblair:
Missed this; do we still need to merge this?

nfultz:
No, someone else (Jasper?) caught it also.

PR #63, opened 2017-11-30
Add citation to two way DLT

nfultz:
Also should we be git'ing vignette/*.R - they get out of sync with .Rmd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,10 @@
 # produced vignettes
 vignettes/*.html
 vignettes/*.pdf
+###vignettes/*.R #TODO NJF probably shouldn't be here
 vignettes/.*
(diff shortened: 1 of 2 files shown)


graemeblair:
Yeah, that seems like a problem. What do you think is best? I guess ideally they are recreated each time we update the Web site via pkgdown.

nfultz:
Notwithstanding pkgdown (eg articles/*.R), I would recommend not keeping the `vignette/*.R` in the repo at all - they get regenerated each time the vignette is knitted.

graemeblair:
Makes sense, thanks. Let's delete & add to .gitignore -- I think they'll still get made by pkgdown we just won't commit them to git.

PR #74, opened 2018-01-08
Nfultz/refactor hof2 cite summary

nfultz:
The main goal of this refactor was to funnel as much NSE through a single place, `declaration_template` - some collateral changes are:

* All delegate functions  now called `delegate`
* All steps can be labeled, labels should be text (**named steps)
* citation functionality is now a step, `declare_citation`
* `design` data structure internals rewritten, list of objects rather than parallel arrays
  * `design$design_function()` gone, use `execute_design(design)`
  * modify_design wrapper not needed, instead use `insert/delete/replace_step` directly.


nfultz:
Future work 
- streamline (PO, Assign, Reveal) triples of steps st reveal is no longer explicitly required in the default case.
- updating vignettes
- purging dead code that is currently commented out

graemeblair:
This is great! I'm not sure we need a user-exposed `execute_design` function, since you can get those from the more useful `get_estimates` and `get_estimands` if I understand correctly. 

Still thinking on better name for `delegate`!

nfultz:
Finally got a green check, merging.

Issue #77, opened 2018-01-15
replace diagnose_design bootstrapped s.e. code with fabricatr resample function


graemeblair:
Issuing this per discussion awhile ago

Issue #53, opened 2017-11-15
drop = FALSE issues in diagnose_design


graemeblair:
When running diagnose_design with a single estimand, you get this:

Error in apply(df[, !colnames(df) %in% group_by_set], 2, sd) : 
  dim(X) must have a positive length

We need to do a package-wide search for drop = FALSE issues like this.

graemeblair:
That line doesn't appear in `diagnose_design` anymore and can't exactly reproduce this. @lukesonnet saw you mentioned this in another PR, do you or @nfultz know if this is still relevant?

nfultz:
Not possible anymore, the SEs are calculated via nested lapply now.

PR #81, opened 2018-01-26
Fix error dropping dim when there is only one diagnosand

lukesonnet:
There used to be an error that when bootstrapping with only 1 diagnosand would throw an error. 

The line I change in R/diagnose_design.R keeps `diagnosands_se_df` as a `data.frame` even if the number of columns not in `group_by_set` is only 1. Otherwise, the `apply` command would fail as df would now be of only one dimension.

I add a test that passes as master currently looks with one diagnosand if you don't bootstrap.

However, the second diagnosis, one diagnosand with bootstrapping, only passes with this branch.

The test also serves as an MWE of the error if you have master currently installed.

### DIFF
--- a/R/diagnose_design.R
+++ b/R/diagnose_design.R
@@ -257,7 +257,7 @@ diagnose_design_single_design <-
         lapply(
           diagnosands_se_df,
           FUN = function(df) {
-            apply(df[, !colnames(df) %in% group_by_set], 2, sd)
+            apply(df[, !colnames(df) %in% group_by_set, drop = FALSE], 2, sd)
           }
         )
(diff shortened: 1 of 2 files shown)


lukesonnet:
Looks like #77 might preclude this and #53 already highlights this. In any case I wanted it for something I was working on so here it is.

nfultz:
Also that chunk of code has been rewritten on my other branch in #78 to allow sorted merging of diagnosand SEs - I'll cherry pick your unit test though (thanks much!).

lukesonnet:
Ugh, I checked your PR first but I checked the wrong file within your PR. Cool, close this when you've picked.

nfultz:
done

Issue #83, opened 2018-01-28
NSE prob


graemeblair:
when using a custom name for your inclusion probs variable, it creates two of them -- seems to be an NSE issue:

```
pop <- declare_population(
  neighborhood = add_level(
    N = 10,
    population = 50
  ),
  citizens = add_level(N = population)
)

smp <- declare_sampling(
  strata = neighborhood, 
  n = 12,
  sampling_variable = S_citizens
)

pop() %>% smp %>% colnames
[1] "neighborhood"              "population"                "citizens"                  "~_inclusion_prob"          "S_citizens_inclusion_prob"
```

Worth checking if this is a problem with assignment variable names too.

graemeblair:
This also causes the subsetting line that drops non-sampled units to introduce NA's into the dataset:

`data[data[, sampling_variable_name] == 1,-which(names(data) %in% sampling_variable_name), drop = FALSE]`

It seems to add rows with all NA's as many as there are sampled observations. Yikes!

PR #86, opened 2018-01-28
Don't propogate NAs through sampling handler

nfultz:
Fixes one part of #83


nfultz:
tested by @graemeblair , pulling into my dev branch.

PR #93, opened 2018-01-31
cleaning up declare_estimator handler

nfultz:
* default handler takes models
* added helper function which can deal with labels
* No longer pass through coefficient_names to models


nfultz:
Merged into nfultz/next

Issue #64, opened 2017-12-01
warning about namespace


graemeblair:
On windows:
*
> diagnosis <- diagnose_design(design, sims = 1000, bootstrap_sims = 50)
Warning messages:
1: In serialize(data, node$con) :
  'package:randomizr' may not be available when loading

PR #113, opened 2018-02-15
Nfultz/foreach to future

nfultz:
Replacing foreach dependencies with future, which seems to work somewhat better on windows - at the minimum, it no longer errors depending on package load ordering, and no longer has serialization warnings. 

This also let us get rid of much of the foreach scaffolding.

side note - collapsed bootstrap/bootstrap_sims arguments into one parameter.


graemeblair:
LGTM, I think the PR error is an estimatr prob?

nfultz:
Mac oldrel failed, I'm just going to restart it and see if it fixes it.

graemeblair:
could this bc mac-oldrel binary doesn't exist on CRAN for future.apply? shouldn't it compile from source if that's the case to install?

nfultz:
Also closes #64

graemeblair:
LGTM

PR #122, opened 2018-03-06
Nfultz/next

nfultz:


graemeblair (commented):
looks great! 

couldn't understand: what is restore()?

nfultz:
restore replaces Z if it was overwritten by fabricate.

graemeblair:
thanks. what's the scenario where `fabricate` overwrites it?

nfultz:
when it's already present. See also alex's ticket about thqt.

graemeblair:
thanks

graemeblair (approved):
LGTM. read the code and understand the restore() solution now. there's a lot of changes here, so suggest we merge when you're ready Neal and then play around with more

graemeblair (commented):
Typo here:
data.frame(coefficient = rownames(summ),

should be coefficients I think

nfultz:
We usually do singular on columns eg estimate

Issue #165, opened 2018-04-12
names / labels steps not recognized


macartan:
This declaration does not keep track of the step label in the summary but it does give user access to the label for replace_step

```
> pop        <- declare_population(N=1)
> design     <- declare_design(pop)
> design

Design Summary

Step 1 (population): declare_population(N = 1) ---------------------------------

new_design <- replace_step(design, pop, declare_population(N=2))

```

This declaration *does* keep track of the label in the summary but it does *not* give user access to the label for replace_step

```
> design     <- declare_design(pop = declare_population(N=1))
> design

Design Summary

Step 1 (population): declare_population(N = 1, label = "pop") ------------------

> new_design <- replace_step(design, pop, declare_population(N=2))
Error in find_step(design, after) : Could not find step (after) in design
```
Desired behavior; (1) Always include label in the summary (2) Always have labels available for replace step etc

graemeblair:
+1 should be possible with `substitute` within `declare_design`

nfultz:
When you use replace, there are three ways of specifying `after`:
  - by object reference (`pop`) as in case 1.
  - by index (`1`) as integer
  - by label (`"pop"`) as text

For your example, the `pop` object is not in the design. You need to quote the label.

> new_design <- replace_step(design, "pop", declare_population(N=2))



Side note - I just spent 5 months getting rid of legacy NSE, please do not reintroduce `substitute` et all.

graemeblair:
Ah I see the difference.

Yes, sorry, just meant NSE. Everyone is on board with rlang.

macartan:
Exactly; so the question is whether the reference can be a label in the
absence of a label and vice versa. Make it easy on people.
Being able to print out a list of the set of steps would help here a lot,
as per top level / print_code requests

nfultz:
Having labels sometimes be labels and sometimes not be labels seems more
confusing to me, not less. How  about we get rid of by-reference mode all
together?

macartan:
No let's keep this functionality, it's useful.
Suggestion is not for labels not to be labels but just for labels to
inherit object reference if no label has been provided.

nfultz:
I don't understand what you mean. The issue is that you posted above is
that the code behaves ambiguously, and the result depends on what objects
you have in you global environment - I think we should eliminate that
ambiguity entirely so that users get predictable code that they can
understand, instead of making it work for one more  use case on the margin
(and interactive!) but which will cause even more confusion down the line.

[...]

PR #169, opened 2018-04-13
Removing find by reference from find_step

nfultz:
per #165, do not accept objects in find_step


nfultz:
@graemeblair

acoppock:
I don't understand if / how this interacts with names and labels

PR #170, opened 2018-04-16
add names in design object to design steps that don't have names

graemeblair:
this makes referring to steps by label easier by ensuring there are labels to each step


nfultz:
I'm mildly opposed to this, becuase the majority of expressions people provide are not valid names - eg `declare_population(N=100)` is not a valid name in R (without backquoting), and it isn't conveying any extra information.

macartan:
This is currently our most common way of declaring designs and it is great:

library(DeclareDesign)
pop <- declare_population(N=2)
smp <- declare_sampling(n = 1)

design <- pop / smp

This is the functionality that would be good in this case:

1 To be able to do
design$pop()

2. To be able to do
  replace_step(design, "pop", pop_2)  or  replace_step(design, pop, pop_2)

Especially as people edit designs around being able to access by name /
label of step will be of value and easier than design[[2]]()

All for any neat approach that let's users do this

nfultz:
@macartan  This PR does not affect building designs with forward slash. Forward slash only appends two designs / steps together.

For `declare_design`, maybe set the names but only if the name would be a valid. I think rlang has a is_symbol that you can use to test that.

nfultz:
Also note that names are not the same as labels, and that replace_step uses
labels only.

PR #174, opened 2018-04-23
Update DESCRIPTION - R 3.4

nfultz:
Updating R requirement, now that 3.5 is out require people to use at least 3.4


graemeblair:
Are there specific features we need?  dplyr depends only on R 3.1.2. Worry this makes users get tired of using it because they have to upgrade.

nfultz:
Yes, in estimatr, fabricatr, randomizr. CRAN will not host current versions of estimatr, fabricatr, randomizr etc for 3.3.

lukesonnet:
You mean because they need to be binaries?

graemeblair:
How does the tidyverse handle this? They do not require the latest version and have the same ecosystem problem.

nfultz:
From a practical standpoint, since cran will not build binaries for DeclareDesign for 3.3 and Travis has been broken for 3.3 mac for a couple months, DeclareDesign transitively requires 3.4 anyway. Especially when it will not work with the 3.3 builds of fabricatr and estimatr. 

If normal people want to keep using 3.3 or older, they should keep using the versions of the packages that were built and tested on 3.3. The people installing packages from source (eg linux users) are savvy enough to force the install or update their R.


There's also a case to just have all users update to 3.5 now, since designs exported from a 3.5 session won't be able to be loaded easily into 3.3/3.4 due to changes in the underlying data structures for saveRDS etc.

acoppock:
Seems like 3.5 is a bigger update? where do we end up on this?

nfultz:
3.4 is the first release that CRAN could provide binaries for mac/win for
DeclareDesign and friends - DeclareDesign won't install install on mac/win
3.3 (eg most users) or lower because CRAN doesn't have binary builds for
randomizr/fabricatr/estimatr of high enough version.

Raising the requirement means that users will get a "not high enough
version of R" error message and fix it themselves, rather than a "can't
install correct version of estimatr/fabricatr/randomizr" error message and
have to figure out that they need to upgrade to get the correct dependency.

If you want to provide binaries (eg drat) for the old version, you will
have to set up travis / appveyor to use the older versions of R and hunt
down the transitive breakage of dependencies (eg nlopt gdal have all been
pain points in the past, testthat and knitr will be different on 3.3, etc)
- and that still doesn't fix the out-of-the-box experience -
`install.packages("DeclareDesign")`  should run cleanly against CRAN and
that will only happen on 3.4 and higher.

acoppock:
thank you -- that's persuasive to me!

graemeblair (approved):

graemeblair:
Thanks Neal

PR #177, opened 2018-05-01
declare_step up

acoppock:
This very small pr implements `declare_step` which has `fabricate` as the default handler. A test confirms it works.


graemeblair (approved):

nfultz:
Steps should all have an inherit tag - 

```
@inheritParams declare_internal_inherit_params
```

acoppock:
thanks will fix!

PR #195, opened 2018-06-02
Append probabilities

### LINKED ISSUE #87: option to add all probs of assignment, not just condition probs
Sometimes you need the other condition probs with complex assignment designs, seems like we could easily add this, i.e. it adds not just Z_cond_prob, but Z_cond_prob_1, Z_cond_prob_0, etc. Thoughts? We'd need to (unless I'm missing something) add a second helper function in randomizr like `obtain_condition_probabilities` that would return the full prob mat given a declaration (right now, there are block_ra_probabilities etc. for each design that can return the full prob mat, but not a function that can take a declaration).

acoppock:
Closes #87

### DIFF
--- a/R/declare_assignment.R
+++ b/R/declare_assignment.R
@@ -53,10 +53,11 @@ declare_assignment <- make_declarations(assignment_handler, "assignment" )
 #' @importFrom rlang quos !!! lang_modify eval_tidy quo f_rhs
 #' @importFrom randomizr conduct_ra obtain_condition_probabilities
 #' @param assignment_variable name for assignment variable
+#' @param append_probabilities_matrix Should the condition probabilities matrix be appended to the data? Defaults to FALSE
 #' @param data a data.frame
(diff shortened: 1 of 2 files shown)


graemeblair (approved):
LGTM

R/declare_assignment.R
```diff
@@ -70,6 +71,13 @@ assignment_handler <-
                         !!cond_prob := obtain_condition_probabilities(!!!options, assignment = !!assn),
        ID_label = NA
       )
+      if(append_probabilities_matrix) {
+        options$N <- quo(nrow(data))
```
nfultz, line 75: When this loops back around to the second assignment variable, N is added to options, so `conduct_ra(N=N, N=nrow(data))` may break since N is NSE. Please either add a test to verify this works for multiple assn variables.

nfultz (commented):
Check that this works with multiple assn variables.

PR #201, opened 2018-06-04
enables get_estimates and get_estimands from multiple designs

### LINKED ISSUE #185: get_estimates() get_estimands() on multiple designs
If we have multiple designs:
```
designs <- expand_design(
  simple_two_arm_designer,
  N = c(10:12))
```
`get_estimates(designs[[1]])` works  but this does not 

```
> get_estimates(designs)
Error in if (step_type %in% names(results)) { : 
  argument is of length zero
```
One can do essentially this using diagnose_design:

```
diagnose_design(designs, sims =1)[[1]]
```
But seems more natural to have get_estimates and get_estimands run on a list.

Note currently wanting this for a metaanalysis design where multiple designs are simultaneously declared and an estimator then takes an estimate from each. 

Aside -- there is a relation here to the diagnose_design(simulations_df...) request: currently diagnose design really does two distinct things that are useful separately (1) run the design one or many times and make a simulations_df (2) calculate diagnosands from a simulations_df;

A structure like this might make this more flexible:

```
run_design <- function(design, sims) {....}
diagnose_design <- function(simulations = NULL, design = NULL, diagnosands, sims = 500,...) {...
  if(is.null(simulations) & is.null(design)) stop("Must specify either a design or a simulations data frame")
  if(is.null(simulations)) simulations <- run_design(design, sims)
  ....
  } 
```

graemeblair:
Works with single design, multiple designs sent to ..., or a list of designs i.e. created from expand_design 

closes https://github.com/DeclareDesign/DeclareDesign/issues/185

### DIFF
--- a/R/design_helper_functions.R
+++ b/R/design_helper_functions.R
@@ -149,23 +149,101 @@ execution_st <- function(design, current_df=NULL, results=NULL, start=1, end=len
 #'
 #' @export
 draw_data <- function(design) {
-  run_design_internal(design, results=list(current_df=0))$current_df
+  run_design_internal(design, results = list(current_df = 0))$current_df
+}
+
+#' #' @rdname post_design
+#' #'
+#' #' @export
+#' get_estimates <- function(design) {
+#'   results <- list("estimator" = vector("list", length(design)))
+#'   run_design_internal.default(design, results = results)$estimates_df
+#' }
+
+#' @rdname post_design
+#'
+#' @export
+get_estimands <- function(...) {
+  
+  designs_quos <- quos(...)
+  designs <- lapply(designs_quos, eval_tidy)
+  
+  ## Two cases:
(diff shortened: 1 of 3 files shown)


acoppock (approved):
LGTM

R/design_helper_functions.R
```diff
@@ -149,23 +149,101 @@ execution_st <- function(design, current_df=NULL, results=NULL, start=1, end=len
...
+  
 }
 
 #' @rdname post_design
 #'
 #' @export
-get_estimands <- function(design) {
-  results=list("estimand"=vector("list", length(design)))
-  run_design_internal.default(design, results=results)$estimands_df
+get_estimates <- function(...) {
```
nfultz, line 204: These should be generics. use get_estimates.design and get_estimates.default

PR #207, opened 2018-06-04
remove slash operator; tests

graemeblair:


R/ops.R
```diff
@@ -1,12 +1,13 @@
 
 
-#' @export
-`/.d_par` <- function(e1,e2) {
-  structure(
-    append(e1, e2),
-    class="design",
-    call=call("/", attr(e1, "call"), attr(e2, "call")))
-}
```
nfultz, line 9: Might as well just delete it, it's in git.

graemeblair: thx

macartan:
I'd like us not to kill this functionality. It is extremely useful for quick creation and modification of designs. Used all the time in the current blog posts. It's simple and clean and helps reduce code.

The below fix declares the design as it goes so runs all checks. "Precocious" warnings arise here in early stages when the full design is not declared. I would much rather cut or limit the precocious warnings than the slash functionality.  Neither of the warnings below are useful.

```
`/.d_par` <- function(e1,e2) {
  
  structure(
    do.call(declare_design, append(e1, e2)),
    class="design",
    call=call("/", attr(e1, "call"), attr(e2, "call")))
}


U  <- declare_population(N = 10, Z = rep(0:1, 5))
Y  <- declare_potential_outcomes( Y ~  Z)
Q  <- declare_estimand(ate = mean(Y_Z_1 - Y_Z_0))

R  <- declare_reveal() 
A  <- declare_estimator(Y ~ Z, estimand = "ate", model = lm_robust, label =  "No control")
  
  
design <- U / Y / Q
Warning messages:
1: Potential outcome is the final step in the design. 
2: Outcome variables (Y) were declared in a Potential Outcome step (), but never later revealed. 
3: Assignment variables (Z) were declared in a Potential Outcome step (), but never later assigned by an assigment step. 

design <- design / R / A
```

graemeblair:
I'm happy to keep exploring and won't drop it; but I'm not yet happy with how it works

macartan:
Here is another approach.
Use / for append; use + for append and declare. 
So + is used in place of / for final element in order to run the appended steps through declare_design

```
`+.d_par` <- function(e1,e2) {
  structure(
    do.call(declare_design, append(e1, e2)),
    class="design",
    call=call("+", attr(e1, "call"), attr(e2, "call")))
}

`/.d_par` <- function(e1,e2) {
  structure(
    append(e1, e2),
    class="design",
    call=call("/", attr(e1, "call"), attr(e2, "call")))
}
```

Then, for example:

```
U  <- declare_population(N = 10, Z = rep(0:1, 5))
Y  <- declare_potential_outcomes( Y ~  Z)
Q  <- declare_estimand(ate = mean(Y_Z_1 - Y_Z_0))
R  <- declare_reveal() 
A  <- declare_estimator(Y ~ Z, estimand = "ate", model = lm_robust, label =  "No control")
  
design <- U / Y / Q / R + A
```

Note issues caught as they should be:

```
> two_arm <- simple_two_arm_design + declare_estimator(Y~Z, model = lm_robust, estimand = "ATE") 
 Error in check_unique_labels(labels, function_types, "estimator") : 
  You have estimators with identical labels: my_estimator
Please provide estimators with unique labels 
```

macartan:
We may want to have a check_design function anyway since users can still create things called designs that have issues. e.g. below using insert_step. Simple function below just washed through declare_design, but one might want to check class and other things. 

```
check_design <- function(design) {do.call(declare_design, design)
                                  return()}

two_arm <- insert_step(two_arm, after = length(two_arm), 
                       declare_estimator(Y~Z, model = lm_robust, estimand = "ATE")) 

> check_design(two_arm)
 Error in check_unique_labels(labels, function_types, "estimator") : 
  You have estimators with identical labels: my_estimator
Please provide estimators with unique labels 

> check_design(simple_two_arm_design)
NULL

graemeblair:
I agree with that issue -- but should we just have the same validation stuff be run at the end of insert_step as well? (and others like it)

macartan:
For sure. All modify type steps should have checks when  poaaible.

check_design should have to be used, only if you are trying to figure out
if there's something wrong with a design. If you wanted to diagnose it, so
to speak.

Am happy with the ///+ solution as a safety belt for the slash declaration

nfultz:
I specifically avoided `+` because it connotes ggplot2-style evaluation -
recommend not mixing + and /.

PR #214, opened 2018-06-05
Expand grid tidy

graemeblair:
This PR allows a function and any other type of object to be used as an `expand_design` argument.


acoppock (approved):
lgtm

PR #216, opened 2018-06-05
add + operator; add wrap_step; make declare_design_internal

graemeblair:
In this pull request, we add the new + operator to replace / and that we suggest be the default way to create designs rather than `declare_design`. Advantages: short, readable, easily add to end of design, syntax is familiar to any user of ggplot2. Tradeoffs: cannot accept any other type of call, so must wrap every step that is mutate etc. in `wrap_step`. See this as a reasonable tradeoff, and actually maybe more understandable for users because *everything gotta be a step*. Also, the design-level tests can't be run only at the end. So they are now run at every step. We handle the fact that this adds `declare_reveal` steps automatically sometimes when there will later be a user-included `declare_reveal` by dropping all the automatically-created ones at each + and re-adding them. This is a little cludgy, but the result is fine (think).


nfultz:
I *strongly* suggest you spend more time thinking about this - wrt to order of operations, etc.  If you had actually wanted to make a full blown DSL then we would have just done that nine months ago.  As it stands, this introduces functionality that does not interact well with other functionality and is not well thought out. Please at least document and test several of the obvious cases  eg `a / b + c /d` and how that is supposed to interact with automatic declare_reveal etc.  

As it stands this idea is not ready to merge.

nfultz:
Also given the paternalistic nature of the rest of the package, I think you are assumming too much here that the average user knows that `+` happens after `/`.  If the users really will be at the level you need to make the rest of the package match that expectation.

graemeblair:
Thanks for weighing in @nfultz. I'm confused -- we're suggesting just one operator, `+`. What are the multiple use cases you're referring to?

nfultz:
If there's only one operator, you should probably use slash instead of plus - otherwise it collides with other things - if there's two you should rethink the path you are going down.

slash has higher precendence than plus, and will work better with ggplot etc.

nfultz:
also if this is the preferred way we should deprecate declare_design and especially the mutate style calls within it.

macartan:
Why not have the design checks just at the end? That's what the a+b+c+end
approach did.

nfultz:
You can't do that with a single operator without heavy NSE, because R (like most languages) evaluates expressions depth-first with some operator precedence tacked on. If you have multiple operators, you may was well just write a full DSL. As it stands this PR introduces quite a lot of untested surface area and more ambiguity for users.  At the minimum all of the newly introduced NSE should be rewritten to use rlang.

You all should choose one way to declare a design, and not have multiple slightly different code paths that yield slightly different answers. Slash sign => append is easy to understand (and easy to understand it's limitations), but plus sign  => partially declare a design is not easy to understand with it doing so many other things. I would like to see way more documentation in this PR to disambiguate what it is supposed to do, a one line man page is not enough.

I would  also suggest that:

```
declare_design(a,b,c,d) 
```

should always result in *exactly* the same thing as

```
a + b + c + d
```

and also because you chose the plus sign

```
x <- a + b
y <- c+ d
d <- x + y
```


As implemented those identities can't hold, at least not without heavy refactoring. 


Also note that 

```
x + y != y + x
```

which is another strike against using plus sign.

Slash (append) never had that problem because it was way less ambitious and never tried to replace declare_design even though some people tried to use it that way. 

Also, either way, wrap_step is very ungainly and redundant given the existence of declare_step. Should choose one and drop the other.

graemeblair:
Thanks @nfultz. Took your suggestion to move all the way to rlang NSE in the operator. 

Can you clarify this part:

> Slash sign => append is easy to understand (and easy to understand it's limitations), but plus sign => partially declare a design is not easy to understand with it doing so many other things.

If we stick with a single operator, as far as I know (and have tested), plus sign fully declares a design.

acoppock:
@nfultz re: the "ungainly" wrap_step().  `declare_step` and `wrap_step` do two different things. `declare_step` is a wrapper for an arbitrary step with a default fabricate handler.  `wrap_step` is (i think!) similar to tidy_estimator in that it takes a function call (i.e., mutate(q = 5) and turns it into a step.  would it help to disambiguate if we called it `tidy_step` ?

nfultz:
> If we stick with a single operator, as far as I know (and have tested), plus sign fully declares a design.

This is the parse tree for `a + b + c + d`:::
```
> codetools::showTree(quote(a+b+c+d))
(+ (+ (+ a b) c) d)
```

each of the intermediate results is a call to declare_design, and only the final one is fully declared. the intermediates could be broken.

> wrap_step is (i think!) similar to tidy_estimator in that it takes a function call (i.e., mutate(q = 5) and turns it into a step.

If you don't need to use wrap_step inside declare_design, why should you have to use it here? We should just make `my_step + mutate(q=5)` work, `my_step + wrap_step(mutate(q=5))` is ungainly even compared to `my_step + declare_step(handler=mutate, q=5)`. This looks like you are letting internal implementation details leak through to the public part of the package.  declare_design already does wrapping, plus sign now uses that code path, so you should just let it handle the wrapping instead of forcing people to do it manually.

R/construct_design.R
```diff
@@ -62,60 +56,110 @@
...
+    
+    ret <- do.call(construct_design, ret)
+    
+  } else {
+    
+    ret <- append(lhs, rhs) 
+    
+    names(ret) <- c(expr_text(enexpr(lhs), nlines = 1), expr_text(enexpr(rhs), nlines = 1))
+    
+    ret <- do.call(construct_design, ret)
```
nfultz: This part looks copy-pasted, needs another pass.

nfultz:
Also steps can't be named with infix notation, should get rid of all the code surrounding that. eg there is no way to do declare_design(foo=declare_population(sleep))

Also need to get rid of the code that special cases a data frame for first step, declare_design(sleep).

R/construct_design.R
nfultz: also use eval_tidy here instead of do.call, otherwise you are forcing the evaluation of rhs too early.

nfultz:
^^^ s/declare_design/construct_design/ as it is now named.

R/construct_design.R
```diff
@@ -62,60 +56,110 @@
...
+  browser()
+  
+  if (!inherits(rhs, "dd")) {
+    stop("The right hand side does not appear to be a dd object. Can you wrap the step with `tidy_step()`?", call. = FALSE)
+  }
+  
+  if (inherits(lhs, "design")) {
+    
+    # browser()
+    auto_steps <- sapply(lhs, function(x) { lbl <- attr(x, "label"); !is.null(lbl) && grepl("Autogenerated by ", lbl) })
```
nfultz: add a subclass or other metadata attribute to the autogenerated steps, don't rely on labels.

R/construct_design.R
graemeblair: thanks, yes, we are working on a tidy version. (it isn't copy-pasted.)

R/construct_design.R
graemeblair: thanks, makes sense, will do

graemeblair:
thanks for these comments, that clarified a lot; we'll try to implement them

acoppock:
Thanks again for pushing us on this @nfultz.

wrt to: 

> We should just make my_step + mutate(q=5) work,

Here's the trouble we're having.

`+.dd` requires dd objects, so we can't seem to put `mutate(q=5)` on the rhs of this operator.

we seem to be able to do this sort of thing with

`%dd%` which doesn't care about object types.

For obvious reasons, we want to stick with `+`.  

In terms of user expectations, it makes a certain amount of sense that you can only pass dd objects to this thing.  Which would require a `tidy_step` or similar.

nfultz:
No, `+.dd` is S3 generic, only the left argument is required to be `.dd` - the right can be whatever. 

```
`+.foo` <- function(e1, e2){
    message("foo", e1, e2)
    
}

a <- structure(0, class="foo")
b <- structure(1, class="foo")
x <- structure(2, class="bar")
a+b
a+x
```

And obviously designs shouldn't start with a dplyr call, (there's no current data to pass them) so it shouldn't be a problem.

acoppock:
gotcha, thanks for that clarification.  How to explain the difference here then?

```
library(dplyr)
library(rlang)

`+.foo` <- function(e1, e2){
  quos(e2)
}

a <- structure(0, class="foo")
b <- structure(1, class="foo")
x <- structure(2, class="bar")
a + b
a + x
a + mutate(q = 5)
```

which returns

```
> a + mutate(q = 5)
Error in mutate_(.data, .dots = compat_as_lazy_dots(...)) : 
  argument ".data" is missing, with no default
```

But this:

```
rm(list = ls())

`+` <- function(e1, e2){
  quos(e2)
}

a <- structure(0, class="foo")
b <- structure(1, class="foo")
x <- structure(2, class="bar")
a + b
a + x
a + mutate(q = 5)
```
returns

```
> a + mutate(q = 5)
[[1]]
<quosure>
  expr: ^e2
  env:  0x10ad670c0
```

puzzling!

graemeblair:
really scratching our heads on this `+.foo` issue, if you have any thoughts @nfultz they would be very much appreciated

nfultz:
@graemeblair https://github.com/DeclareDesign/DeclareDesign/blob/master/R/declare_reveal.R#L157

PR #231, opened 2018-06-18
capitalization issue fix

### LINKED ISSUE #221: diagnosis nice names
It is nice that the reshape_diagnosis does spacing and capitalization but it can be confusing that it also does it with the parameter names. eg if I have two parameters, m and M, they will both appear as M in the diagnosis output; better to leave the parameters out of the text reformatting step

graemeblair:
closes https://github.com/DeclareDesign/DeclareDesign/issues/221

changes diagnose_design so it always returns the parameters_df (this seems useful for more reasons than this and cheap)

### DIFF
--- a/R/diagnosis_helper_functions.R
+++ b/R/diagnosis_helper_functions.R
@@ -168,11 +168,13 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
   return_df$design_label <- factor(return_df$design_label, levels = c(levels(return_df$design_label), ""))
   return_df[return_df$statistic == "SE (bootstrapped)", sort_by_list] <- ""
   
+  parameter_names <- names(diagnosis$parameters_df)[-1]
+  
   # Make names nicer
-  names(return_df) <-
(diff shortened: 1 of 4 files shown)


acoppock (approved):
lgtm

R/diagnosis_helper_functions.R
```diff
@@ -168,11 +168,13 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
   return_df$design_label <- factor(return_df$design_label, levels = c(levels(return_df$design_label), ""))
   return_df[return_df$statistic == "SE (bootstrapped)", sort_by_list] <- ""
   
+  parameter_names <- names(diagnosis$parameters_df)[-1]
+  
   # Make names nicer
-  names(return_df) <-
+  names(return_df)[which(!names(return_df) %in% parameter_names)] <-
```
nfultz, line 174: use setdiff

graemeblair: thanks will fix

PR #232, opened 2018-06-18
remove out-dated references to declare_design

### LINKED ISSUE #228: diagnose_design check for declare has inappropriate error
The error in these lines should be rewritten given that `declare_design` is deprecated. Also, if you pass `parallel = TRUE` this causes an error, but I guess that's because that's no longer an argument and it goes into the dots instead.

https://github.com/DeclareDesign/DeclareDesign/blob/a01ecce67b21acf89f491b39bd22a71b7fe4cd9b/R/diagnose_design.R#L107

graemeblair:
closes https://github.com/DeclareDesign/DeclareDesign/issues/228

### DIFF
--- a/R/construct_design.R
+++ b/R/construct_design.R
@@ -15,7 +15,7 @@
 #'
 #' The location of the estimand and estimator functions in the pipeline of functions determine *when* the values of the estimand and estimator are calculated. This allows users to, for example, differentiate between a population average treatment effect and a sample average treatment effect by placing the estimand function before or after sampling.
 #'
-#' Designs declared with declare_design can be investigated with a series of post-declaration commands, such as \code{\link{draw_data}}, \code{\link{get_estimands}}, \code{\link{get_estimates}}, and \code{\link{diagnose_design}}.
+#' Design objects declared with the + operator can be investigated with a series of post-declaration commands, such as \code{\link{draw_data}}, \code{\link{get_estimands}}, \code{\link{get_estimates}}, and \code{\link{diagnose_design}}.
 #'
 #' The print and summary methods for a design object return some helpful descriptions of the steps in your research design. If randomizr functions are used for any assignment or sampling steps, additional details about those steps are provided.
 #'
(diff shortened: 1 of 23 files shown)


acoppock (approved):
ok

R/construct_design.R
```diff
@@ -15,7 +15,7 @@
 #'
 #' The location of the estimand and estimator functions in the pipeline of functions determine *when* the values of the estimand and estimator are calculated. This allows users to, for example, differentiate between a population average treatment effect and a sample average treatment effect by placing the estimand function before or after sampling.
 #'
-#' Designs declared with declare_design can be investigated with a series of post-declaration commands, such as \code{\link{draw_data}}, \code{\link{get_estimands}}, \code{\link{get_estimates}}, and \code{\link{diagnose_design}}.
+#' Design objects declared with the + operator can be investigated with a series of post-declaration commands, such as \code{\link{draw_data}}, \code{\link{get_estimands}}, \code{\link{get_estimates}}, and \code{\link{diagnose_design}}.
```
nfultz, line 18: suggest changing this from "are declared" to "are created by linking together declared steps" or similar.

R/design_helper_functions.R
```diff
@@ -248,7 +248,7 @@ get_estimands_single_design <- function(design) {
 
 #' Obtain the preferred citation for a design
 #'
-#' @param design a design object created by \code{declare_design}
+#' @param design a design object created using the + operator
```
nfultz, line 251: this is awkward phrasing, what does hadley use in ggplot2 docs?

Also can use inheritparams to make sure its consistent.

nfultz (approved):
Seems fine, doc only changes.  Hopefully this fixes the build so that examples all run correctly.

graemeblair:
thanks @nfultz merging but opened https://github.com/DeclareDesign/DeclareDesign/issues/237 to more consistently change things across the package along the lines you suggest

PR #240, opened 2018-06-19
enable functions to simulate_design

### LINKED ISSUE #230: simulate a design based on a custom design function rather than design object
@macartan suggests allowing users to write their own design function that returns a run of the design (i.e. estimates and estimands)

This seems in line with our general philosophy of enter-and-exit-our-framework-at-will. Does not seem too hard to implement either. 

The function they write would have to return a list with two entries, estimates and estimands.

Others, thoughts?

graemeblair:
closes https://github.com/DeclareDesign/DeclareDesign/issues/230

you can run, simulate, and diagnose functions that return lists of estimates_df and estimands_df


R/diagnose_design.R
```diff
@@ -104,10 +104,10 @@ diagnose_design <- function(...,
     } else {
       designs <- dots
     }
-    if (!all(vapply(designs, inherits, FALSE, "design"))) {
-      stop(
-        "Please only send design objects created using the + operator to diagnose_design."
-      )
+    if (!all(sapply(designs, function(x) {
```
nfultz: pull this into a helper function.

R/simulate_design.R
```diff
@@ -96,11 +99,8 @@ simulate_design <-
       SIMPLIFY = FALSE
     )
     
-    #if (length(designs) > 1) {
-      simulations_list <-
-        Map(cbind, design_label = names(simulations_list), simulations_list, stringsAsFactors = FALSE)
-    #}
-    
+    simulations_list <-
```
nfultz: join these lines

R/diagnose_design.R
nfultz: and also use it in simulate

R/design_helper_functions.R
```diff
@@ -127,6 +127,14 @@ run_design_internal.default <- function(design, current_df=NULL, results=NULL, s
   
 }
 
+# for when the user sends a function that runs a design itself 
+#   to run_design (or simulate_design / diagnose_design above it)
+run_design_internal.function <- function(design) {
```
nfultz: should also update the generic on run_design_internal.default to be design, now that accept multiple types, and have default thow an error.

nfultz (dismissed):
seems fine.

graemeblair:
thx, addressed

PR #242, opened 2018-06-19
Addresses #234

### LINKED ISSUE #234: More reshape details
In reshape_diagnosands: Currently the select columns code block is before the name cleaning and so select error message (ie use "bias", "estimand_label" etc) uses the wrong names; also select not being applied at all for cases with no bootstrapping; also cleaning code is there twice.

Order that was there before should be fine:

```
 if (is.null(diagnosis$bootstrap_replicates))  {
  out <- diagnosands_df
  } else {
  out <- reshaping
  }

# Then:
nice_names(out)
# Then:
select(out)
```

Other things:

1. parameters are still being capitalized
2. Too much repetition in se rows: we do not need repetition of parameter names or of N_sims (I also think we we do not need se (bootstrapped) unless there is some other possibility)
3. select should throw an error when only one column is selected

acoppock:
This PR makes it so the select operates on the nice names. closes #234

### DIFF
--- a/R/diagnosis_helper_functions.R
+++ b/R/diagnosis_helper_functions.R
@@ -122,7 +122,7 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
   # Reshape if there is bootstrapping
   diagnosand_columns <- diagnosis$diagnosand_names
   diagnosand_se_columns <- paste0("se(", diagnosis$diagnosand_names, ")")
-  group_columns <- names(diagnosands_df)[!names(diagnosands_df) %in% c(diagnosand_columns, diagnosand_se_columns)]
+  group_columns <- setdiff(names(diagnosands_df), c(diagnosand_columns, diagnosand_se_columns))
   
   # Make diagnosand only df
   diagnosands_only_df <-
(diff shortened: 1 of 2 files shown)


graemeblair (dismissed):
LGTM, let's wait for @nfultz review to merge in

R/diagnosis_helper_functions.R
```diff
@@ -154,30 +154,38 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
...
-  rownames(return_df) <- NULL
+  }
   
+  names_to_change <- setdiff(names(return_df), parameter_names)
+  names(return_df)[names(return_df) %in% names_to_change] <- make_nice_names(names_to_change)
+    
+  # Select columns
+  if (!is.null(select)) {
+    available_to_select <- make_nice_names(c(group_columns, diagnosand_columns))
+    if (any(!(select %in% available_to_select)))
```
nfultz: de morgans law => !all

R/diagnosis_helper_functions.R
```diff
@@ -154,30 +154,38 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
...
+      stop(paste(
+        "select argument must only include elements from: ",
+        paste(available_to_select, collapse = ", ")
+      ))
```
nfultz: add a test to trigger this stop.

nfultz (dismissed):
lgtm

acoppock:
Thanks, those changes in.

graemeblair (approved):
good2go

PR #246, opened 2018-06-21
names labels

### LINKED ISSUE #220: add name to step when auto injecting reveal step

### LINKED ISSUE #227: Better default names
If the convention is going to be `U + Y + Z + Q + S + R + A` to build designs, we should build better default names for anonymous steps. eg

```
declare_population(N=100, foo=1) + declare_sampling(m=50)
```

should be named something like U and S instead of being named after the deparsed call. Perhaps it could be letter dot dot position.

For each step type, declare a default letter, and use that to make names that look like U..1, S..2, etc.

acoppock:
closes #220 
closes #227 
closes #246


R/construct_design.R
```diff
@@ -82,41 +82,32 @@
...
     if (!inherits(rhs, "dd") && !inherits(rhs, "function")) {
       stop("The right hand side does not appear to be a DeclareDesign object or a function", call. = FALSE)
     }
     
     if (inherits(lhs, "design")) {
       
-     # auto_steps <- sapply(lhs, function(x) { lbl <- attr(x, "label"); !is.null(lbl) && grepl("Autogenerated by ", lbl) })
-     auto_steps <- sapply(lhs, function(x) !is_null(attr(x, "auto-generated")))
-        
+      auto_steps <- sapply(lhs, function(x) !is_null(attr(x, "auto-generated")))
```
nfultz: pull out predicate to helper file, rewrite as Filter(Negate(is_autogenerated), lhs)

R/construct_design.R
```diff
@@ -128,31 +119,19 @@
 construct_design <- function(...) {
   
   qs <- quos(...)
+  qs <- maybe_add_names_qs(qs)
```
nfultz: note that you don't really use qnames now, can probably purge this and set names(ret) directly.

R/construct_design.R
```diff
@@ -192,56 +187,6 @@ construct_design <- function(...) {
...
-  
   ret
 }
 
-###############################################################################
-# In a design created using the + operator, if a step is a dplyr style call mutate(foo = bar),
-# it will fail to evaluate - we can catch that and try to curry it
-
-#' @importFrom rlang quos lang_fn lang_modify eval_tidy
-callquos_to_step <- function(step_call, label="") {
```
nfultz, line 212: +1

R/declare_potential_outcomes.R
```diff
@@ -290,54 +290,32 @@ pofdv <- function(design, i, step){
...
-    new_step <- eval_tidy(quo(declare_reveal(outcome_variables = !!this_step_meta$outcome_variables,
-                                             assignment_variables = !!this_step_meta$assignment_variables,
-                                             label = !!paste("Autogenerated by", attr(step, "label"))) ))
-    attr(new_step, "auto-generated") <- TRUE
-    
-    for (step_j in design[length(design):(i + 1)]) {
-      if (is.null(step_meta <- attr(step_j, "step_meta"))) next;
-      if (attr(step_j, "step_type") == "assignment") {
-        if (any(step_meta$assignment_variables %in% attr(step, "step_meta")$assignment_variables)) {
-          # warning("Attempting to inject a `declare_reveal(", this_step_meta$outcome_variables, ", ",
```
nfultz, line 313: The warnings should probably come back at some point.

R/utilities.R
```diff
@@ -5,35 +5,53 @@
...
-    quotations[[i]] <- labeller(quotations[[i]], names(quotations)[i])
-  }
-  
-  if (any(duplicated(names(quotations)))) {
-    stop(paste0("Please provide unique names for each design step. Duplicates include ", 
-                paste(names(quotations)[duplicated(names(quotations))], collapse = ", "), 
-                ". You can name steps within declare design, e.g. declare_design(my_pop = pop_step)."), call. = FALSE)
+    if (names(quotations)[i] == ""){
+      names(quotations)[i] <- namer(quotations[[i]])
+    }
```
nfultz, line 38: whole loop can be vectorized now.

nfultz (changes requested):

R/construct_design.R
graemeblair: thx. couldn't get this to work. tried: 

is_autogenerated <- function(steps) {
  sapply(steps, attr, "auto-generated")
}

lhs <- Filter(Negate(is_autogenerated), lhs)

am I missing something?

R/construct_design.R
graemeblair: we do that a bit lower, but need here to capture names like pop when sent to `pop + declare_sampling(n = 50)`. the name `pop` is lost if you don't use maybe_add_names on the quosures.

R/declare_potential_outcomes.R
graemeblair: thx, yes, will remember to come back to this

nfultz:
See https://github.com/DeclareDesign/DeclareDesign/pull/249

R/construct_design.R
nfultz: There's only one place that calls in to conduct_design, should not need any NSE within it - pass the steps directly. See #249

tests/testthat/test-fanout.R
```diff
@@ -84,6 +84,7 @@ test_that("sims expansion is correct",{
 
 
 test_that("fan_out ids are correct",{
+  sleep_large <- resample_data(sleep, N = 200)
```
nfultz: is this used anywhere?

R/construct_design.R
```diff
@@ -82,56 +82,56 @@
...
+  
+  for(i in match(to_rename, names(rhs))) {
+    if(attr(rhs, "valid")) next;
+    new_nm <- names(rhs)[i]
     
-    if (!inherits(rhs, "dd") && !inherits(rhs, "function")) {
-      stop("The right hand side does not appear to be a DeclareDesign object or a function", call. = FALSE)
-    }
+    while(new_nm %in% union(names(lhs), names(rhs)))
+      new_nm <- paste0(new_nm, "._")
```
nfultz: Needs to add numbers like data.frame does.

nfultz (changes requested):

tests/testthat/test-fanout.R
graemeblair: that test fails intermittently, was trying to see if upping sample size would help

R/construct_design.R
graemeblair: great, just made commits along these lines

graemeblair:
think this is good to go, subject to @nfultz review

tests/testthat/test-fanout.R
graemeblair: oic I didn't actually implement. commit to follow

PR #249, opened 2018-06-21
refactoring nse in conduct_design -> plus

nfultz:
consolidates NSE in plus sign


graemeblair (approved):
Thanks for doing this @nfultz, much appreciated LGTM.

PR #252, opened 2018-06-21
bug fixes for multiple assignment vars

### LINKED ISSUE #236: Error on misisng conditions
Error not very informative:
```
> Y  <- declare_potential_outcomes(Y ~ Z1 + Z2, assignment_variables = list("Z1", "Z2"))
Error in names(object) <- nm : 
  'names' attribute [2] must be the same length as the vector [1]
```
This is correct specification:

```
 Y  <- declare_potential_outcomes(Y ~ Z1 + Z2, conditions = list(Z1 = 0:1, Z2 = 0:1))
```

Assuming binary treatments  might be reasonable behavior when only variables are specified

graemeblair:
closes https://github.com/DeclareDesign/DeclareDesign/issues/236

you can now declare multiple assignment variables via

`declare_potential_outcomes(..., conditions = list(Z1 = 0:1, Z2 = 1:2))` OR
`declare_potential_outcomes(..., assignment_variables = c("Z1", "Z2"))` which defaults to conditions c(0, 1)
`declare_potential_outcomes(..., assignment_variables = list("Z1", "Z2"))` which also defaults to conditions c(0, 1)

### DIFF
--- a/R/declare_potential_outcomes.R
+++ b/R/declare_potential_outcomes.R
@@ -183,10 +183,10 @@ potential_outcomes.formula <- function(formula,
 
 
 validation_fn(potential_outcomes.formula) <- function(ret, dots, label) {
+  
   dots$formula <- eval_tidy(dots$formula)
   outcome_variable <- as.character(dots$formula[[2]])
-  
-  
+
   if (length(dots$formula) < 3) {
     declare_time_error("Must provide an outcome in potential outcomes formula", ret)
   }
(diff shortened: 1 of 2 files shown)


nfultz (approved):
lgtm

nfultz:
I still think we should deprecate assignment_variables, and just have people specify conditions only.

graemeblair:
thx. I'm ok with that I think, but suggest we do post-CRAN

PR #254, opened 2018-06-22
add_citation

acoppock:
removes declare_citation, adds add_citation


nfultz (changes requested):
rename it set_citation

Issue #259, opened 2018-06-22
Difficulty appending steps / missing names


macartan:
```
library(DesignLibrary)

processing <- declare_step(handler = fabricate,
                           neighbor = sapply(1:N, function(i) sum(Z[-i][G[-i]==G[i]])),
                           one_neighbor  = as.numeric(neighbor==1),
                           two_neighbors = as.numeric(neighbor==2))

linear     <- declare_estimator(Y ~ Z*neighbor, estimand = "Treat_1", model = lm_robust, 
                                label = "linear")

saturated  <- declare_estimator(Y ~ Z*one_neighbor + Z*two_neighbors, estimand = "Treat_1", 
                                label = "saturated", model = lm_robust)

neighbors_design <- insert_step(simple_spillover_designer(), after = 3, processing) 

# Issue 1: inserted step has no name
names(neighbors_design)

# Issue 2: append difficulties

check0 <- neighbors_design + linear 
check1 <- neighbors_design + saturated 
check2 <- neighbors_design + linear + saturated
```

1. inserted step did not create a name
2. check2 fails

```
> # Issue 1: inserted step has no name
> names(neighbors_design)
[1] "population" "estimand"   "assignment" ""           "reveal"     "estimator" 
> 
> # Issue 2: append difficulties
> 
> check0 <- neighbors_design + linear 
> check1 <- neighbors_design + saturated 
> check2 <- neighbors_design + linear   + saturated
Error in names(ret)[qnames != ""] <- qnames[qnames != ""] : 
  NAs are not allowed in subscripted assignmentsNAs are not allowed in subscripted assignments
```

graemeblair:
yikes. the two are likely connected. will fix.

graemeblair:
closed by #260

PR #260, opened 2018-06-22
Insert modify fix

acoppock:
This PR updates the modify / insert / delete / add infrastructure to rlang. Addresses #259 naming troubles.


nfultz:
Why is NSE needed here at all?

graemeblair:
the messages were all wrong -- telling you which step could not be found

R/modify_design.R
```diff
@@ -56,27 +58,26 @@ NULL
...
+insert_step <- function(design, new_step, before, after) {
+  insert_step_(design, enquo(new_step), enquo(before), enquo(after))
 }
 
-insert_step_ <- function(design, new_step_quosure, before = NULL, after = NULL) {
-  if (is.null(after)) {
-    if (is.null(before)) {
+insert_step_ <- function(design, new_step_quosure, before, after) {
+  if (quo_is_missing(after)) {
+    if (quo_is_missing(before)) {
```
nfultz: Keep the nulls here, just so people that are trying to program against this get correct error eg when

insert_step(design, auto_reveal, after=design$assignment_step)

if after is accidentally null, still want this error, not a downstream, less clear one.

R/modify_design.R
```diff
@@ -56,27 +58,26 @@ NULL
...
       stop("Must provide either before or after to add_step()")
     }
     after <- find_step(design, before) - 1
   } else {
     after <- find_step(design, after)
   }
   
-  new_step <- eval_tidy(new_step_quosure) 
+  new_step <- wrap_step(eval_tidy(new_step_quosure), f_rhs(new_step_quosure))
```
nfultz: instead of calling construct_design directly, you should be able to do something like

design <- eval_tidy( design[i <= after] + !!new_step_quosure + design[i > after] )

R/modify_design.R
```diff
@@ -1,14 +1,16 @@
 # index of a step (specified by object, label or position)
 find_step <- function(design, step) {
-  if (is.numeric(step) && step <= length(design) && step > 0) return(step)
-  if (is.character(step)) {
+  step_obj <- eval_tidy(step)
```
nfultz:
```
find_step <- function(design, step_quo)
   step <- eval_tidy(step_quo)
```

quosures are the unusual things, not steps, so try to name variables according to level of surprise.

nfultz: But I think find step can be rewritten so that all these quos are not needed.

nfultz (changes requested):

nfultz:
Do we really need to echo back which step couldn't be found?

eg is

```
>insert_step(design, new_step, old_step)
Couldn't find old_step
```

really that much more helpful than

```
>insert_step(design, new_step, old_step)
Couldn't locate position in design
```

For pretty printing error messages, we ought to use tryCatch and not try to thread through four layers of NSE.

R/modify_design.R
graemeblair: couldn't get this to work because design[i <= after] is not a design, it's a list, so + didn't call +.dd

R/modify_design.R
graemeblair: fixed thx

nfultz:
minor but #265

PR #270, opened 2018-06-23
Nfultz/simulation cleanup

nfultz:
removing copied code from get_estimates/estimands/simulate/diagnose, makes helper function instead.

as a consequence, all of theoe now also work with anything that has a run_design_internal method (eg functions with zero args), all though many combinations are not tested at this point.


graemeblair (approved):
lgtm

PR #275, opened 2018-06-25
reshape issues fixed

graemeblair:
- fixed renaming of parameters (capitalizing)
- fixes select so it works whether you bootstrap or not

### DIFF
--- a/R/diagnosis_helper_functions.R
+++ b/R/diagnosis_helper_functions.R
@@ -106,57 +106,63 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
   
   diagnosands_df <- diagnosis$diagnosands
   
-  # If no bootstrapping
-  if (is.null(diagnosis$bootstrap_replicates)) {
-    # Make names nicer
-    names(diagnosands_df) <-
-      gsub("\\b(se[(]|sd |rmse|[[:alpha:]])",
-           "\\U\\1",
-           gsub("_", " ", names(diagnosands_df)),
-           perl = TRUE)
-    rownames(diagnosands_df) <- NULL
+  # # If no bootstrapping
+  # if (is.null(diagnosis$bootstrap_replicates)) {
+  #   # Make names nicer
+  #   names(diagnosands_df) <-
+  #     gsub("\\b(se[(]|sd |rmse|[[:alpha:]])",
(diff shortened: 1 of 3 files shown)


R/diagnosis_helper_functions.R
```diff
@@ -77,7 +77,7 @@ print.summary.diagnosis <- function(x, ...) {
   n_sims <- unique(x$diagnosands_df$n_sims)
   cat(paste0("\nResearch design diagnosis", ifelse(length(n_sims) == 1, paste0(" based on ", n_sims, " simulations"), ""), "."))
   if (x$bootstrap_sims > 0) {
-    cat(" Diagnosand estimates with bootstrapped standard errors (", x$bootstrap_sims, " replicates).", sep = "")
+    cat(" Diagnosand estimates with bootstrapped standard errors in parentheses (", x$bootstrap_sims, " replicates).", sep = "")
```
nfultz, line 80: should use message or warning instead of cat.

R/diagnosis_helper_functions.R
```diff
@@ -106,57 +106,52 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
...
-  # If no bootstrapping
-  if (is.null(diagnosis$bootstrap_replicates)) {
-    # Make names nicer
-    names(diagnosands_df) <-
-      gsub("\\b(se[(]|sd |rmse|[[:alpha:]])",
-           "\\U\\1",
-           gsub("_", " ", names(diagnosands_df)),
-           perl = TRUE)
-    rownames(diagnosands_df) <- NULL
+  if(!is.null(diagnosis$bootstrap_replicates)){
```
nfultz: is.data.frame

nfultz: This is a very large if/else block - it was easier to read when the small edge case was close to the if statement, instead of 50 lines away. Probably should refactor this large block to be clean_bootstrap_replicates or similar - use rstudio's extract function command.

nfultz (dismissed):
seems ok

R/diagnosis_helper_functions.R
graemeblair: ok thanks will do

jaspercooper:
Nice

PR #284, opened 2018-07-09
broom interoperability

### LINKED ISSUE #278: does confint need to be a data.frame in estimator_handler?
Right now `estimator_handler` is choking on a `confint()` object returned by `systemfit` because it is a matrix classed as `confint.systemfit` and therefore can't be coerced to a data.frame. Seems a shame because really we are quite close to this working. It makes me wonder whether other estimators might exhibit this issue
MWE:
```r
library(DeclareDesign)
library(systemfit)
library(DesignLibrary)
data <- draw_data(crossover_designer())
sur_fit <- systemfit(
			formula = list(YA ~ A, YB ~ B),
			method = "SUR",
			data = data)
SUR <- declare_estimator(
  formula = list(YA ~ A, YB ~ B),
  model = systemfit,
  method = "SUR",
  coefficients = "eq1_A",
  label = "SUR"
)
# Looks great:
as.matrix(confint(sur_fit))
# Awesome, we can pull the confint:
confint(sur_fit)["eq1_A",]
# Womp:
SUR(data)
as.data.frame(confint(sur_fit))
```

acoppock:
This PR allows us to partially depend on broom to handle many many kinds of model objects.  One growing pain: we have to adopt broom's column names:
term
estimate
std.error
conf.high
conf.low

etc.

This means people can now really use most any modeling function.

closes https://github.com/DeclareDesign/DeclareDesign/issues/278


tests/testthat/test-bootstrap-diagnosands.R
```diff
@@ -32,6 +32,6 @@ test_that("test diagnosands", {
 
   expect_equal(dim(diagnosis$diagnosands_df), c(2,23))
 
-  expect_equal(dim(diagnosis$simulations_df), c(4,11))
+  expect_equal(dim(diagnosis$simulations_df), c(4,14))
```
nfultz, line 35: ?

R/declare_estimator.R
```diff
@@ -261,44 +292,65 @@ validation_fn(model_handler) <-  function(ret, dots, label){
...
-               summ,
-               ci,
-               stringsAsFactors = FALSE, row.names = NULL)
-  colnames(return_data) <- c("coefficient","est", "se", "p", "ci_lower", "ci_upper")
-
-
-  if (is.character(coefficients)) {
-    coefs_in_output <- coefficients %in% return_data$coefficient
+fit2tidy <- function(fit, term = FALSE) {
+  if (requireNamespace("broom", quietly = TRUE)) {
```
nfultz, line 296: should pull this chunk out into methods, similar to estimatr's fake broom https://github.com/DeclareDesign/estimatr/blob/master/R/S3_tidy.R.

nfultz (dismissed):
Overall I like the words tidy_df / terms much more than return_data / coefficients.

tests/testthat/test-bootstrap-diagnosands.R
acoppock: now we return whatever broom returns, so the simulationsdf has more columns!

R/declare_estimator.R
acoppock: sorry, I don't understand -- this works on any fit object, so how can it be a method?

graemeblair: we need to create s3 methods for each function we want to support as we do in estimatr, and rely on the already existing ones from estimatr, via UseMethod("tidy") instead of running directly in fit2tidy

graemeblair: I think we can't do this, because the main thing we'd need to do is tidy.lm and tidy.glm - which would collide with those methods in broom. Do you agree @nfultz?

nfultz: by implementing DeclareDesign::tidy.default and masking estimatr::tidy.default, I think this would work. No need to reinvent tidy.lm or tidy.glm.

graemeblair: ok thanks we'll try that

PR #285, opened 2018-07-10
add declare time error if >1 label sent

### LINKED ISSUE #280: Names error when multiple labels given in declare_estimator
Full example below but here is the key thing:

I wanted to give two lables in declare_estimator so that these would be distinct estimator labels:

```
> estimator_1 <- declare_estimator(Y ~ A + B,
+                                      fixed_effects = ~ blocks,
+                                      model = lm_robust,
+                                      coefficients = c("A", "B"),
+                                      estimand = c("ate_A", "ate_B"), label = c("est_A", "est_B"))
> estimator_1(df)
  estimator_label coefficient       est        se         p    ci_lower  ci_upper estimand_label
1           est_A           A 0.2408310 0.1797732 0.1835257 -0.11601593 0.5976779          ate_A
2           est_B           B 0.2768489 0.1799845 0.1272938 -0.08041756 0.6341153          ate_B
```
But when I declare this throws an error

```
> simple_factorial_design <- population + potentials + 
+                                estimand_1 + estimand_2 + estimand_3 +
+                                assign_A + assign_B + reveal + 
+                                estimator_1
 Error in names(object) <- nm : 
  'names' attribute [2] must be the same length as the vector [1] 
```

In contrast if I use one label I can declare

```
> estimator_1 <- declare_estimator(Y ~ A + B,
+                                      fixed_effects = ~ blocks,
+                                      model = lm_robust,
+                                      coefficients = c("A", "B"),
+                                      estimand = c("ate_A", "ate_B"), label = c("est_A"))
> simple_factorial_design <- population + potentials + 
+                                estimand_1 + estimand_2 + estimand_3 +
+                                assign_A + assign_B + reveal + 
+                                estimator_1
> get_estimates(simple_factorial_design)
  estimator_label coefficient         est        se         p   ci_lower  ci_upper estimand_label
1           est_A           A -0.07073160 0.1791042 0.6937790 -0.4262507 0.2847875          ate_A
2           est_A           B  0.03438051 0.1789219 0.8480277 -0.3207767 0.3895377          ate_B
```
Not thrilled to have the same estimator label but there is no mixing up of estimates because the coeffcient is different

* Do we want this behavior
* If so can we throw a more informative error ("labels of length 1 only")

graemeblair:
closes #280


nfultz (changes requested):
Instead of adding this to every single step validator, you should put in validate()

graemeblair:
thanks, tried that in latest push -- doesn't run that code for any step so doesn't throw error. is what I did here what you were suggesting?

R/aaa.R
```diff
@@ -175,6 +175,10 @@ has_validation_fn <- function(f){
 }
 
 validate <- function(handler, ret, dots, label) {
+  # error if label is length > 1
+  if(is.character(label) && length(label) > 1)
+    declare_time_error("Please provide only one label.", declaration)
```
nfultz: s/declaration/ret/

nfultz (dismissed):
need to fix the error msg but ow lgtm

R/aaa.R
graemeblair: thanks corrected. 

but validate() isn't being called for some reason on any step (or at least that code is not being run) i.e.:

`declare_population(N = 5, label = c("lbl1", "lbl2"))`

does not result in error

graemeblair:
@nfultz if you have any thoughts on why this isn't working -- I moved it to validate as suggested, but that code isn't run in the tests that worked when it was done in the individual declare_* validation functions.

nfultz:
There might be some older logic around only calling validate if a validation_fn is set on the handler - that's no longer the case since all handlers can only take 1 label;.

graemeblair:
thx, will check it out

graemeblair:
closing this PR; got it working in another

PR #286, opened 2018-07-10
Fan out syntax

### LINKED ISSUE #266: fan_out syntax
In one of the tests, we have:

```
  sleep_large <- resample_data(sleep, N = 200)
  design <- declare_population(sleep_large) +
    declare_estimand(2, label = "a") +
    declare_estimand(b = rnorm(1))
  
  fan <- data.frame(end = 1:3, n = c(5, 5, 5))

  fo <- diagnose_design(design, sims = fan)
```

This seems like it's testing internal functionality, but it's causing trouble in `simulate_design`.  @nfultz , are we correct in thinking that user will only ever pass a vector to `sims` and not a data.frame?

graemeblair:
unexports execution_st and errors if you try to send an execution step to either diagnose_design or simulate_design. leaves all execution step functionality for internal use.

closes https://github.com/DeclareDesign/DeclareDesign/issues/266


R/design_helper_functions.R
```diff
@@ -141,15 +141,6 @@ run_design_internal.function <- function(design) {
...
 
-#' Build an execution strategy object
-#'
-#' @param design a design
-#' @param current_df a data.frame
-#' @param results a list of intermediate results
-#' @param start index of starting step
-#' @param end  index of ending step
-#'
-#' @export
```
nfultz, line 152: leave the comments in, just remove the apostrophes to get roxygen to quit.

R/simulate_design.R
```diff
@@ -49,6 +49,9 @@ simulate_design <- function(..., sims = 500) {
     
     designs <- dots_to_list_of_designs(...)
     
+    if (!is.numeric(sims) & !is.vector(sims)) {
```
nfultz: use rlang helper here instead.  is_bare_numeric.

nfultz: actually is_bare_intergerish would be better.

nfultz (dismissed):
seems fine.

nfultz (dismissed):
LGTM

PR #296, opened 2018-07-11
Reshape design digits fix

graemeblair:
fixes case where there is no bootstrapping, now digits works


R/diagnosis_helper_functions.R
```diff
@@ -120,6 +120,11 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL) {
   } else {
     group_columns <- setdiff(names(diagnosands_df), diagnosand_columns)
     return_df <- diagnosands_df
+    return_df[, diagnosand_columns] <-
+      data.frame(lapply(return_df[, diagnosand_columns, drop = FALSE],
```
nfultz: You shouldn't need to convert to data.frame here - you can assign a list into a set of columns directly.
```
return_df[diag_cols] <- lapply(return_df[diag_cols], format_num, digits=digits)
```

PR #310, opened 2018-08-16
deal with NAs in group_by_set

### LINKED ISSUE #301: disappearing diagnosis
I have a design that produces the following simulations df using simulate_design 

```
> sims_df
  design_label sim_ID estimand_label    estimand estimator_label         est coefficient        se         p   ci_lower  ci_upper
1       design      1        control -0.06675827              A1 -0.06334958        <NA>        NA        NA         NA        NA
2       design      1            ate  0.17865929              A2  0.13914128           Z 0.2603802 0.5944165 -0.3782478 0.6565304
> diagnose_design(sims_df)
```

Note the <.NA.> for coefficient for A1. ( A1 uses a custom function and A2 uses Y~Z.)

There should be two rows to the diagnosis, but there is only 1:

```
Research design diagnosis based on 1 simulations. Diagnosand estimates with bootstrapped standard errors in parentheses (100 replicates).

 Design Label Estimand Label Estimator Label Coefficient N Sims   Bias   RMSE  Power Coverage Mean Estimate SD Estimate Mean Se
       design            ate              A2           Z      1  -0.04   0.04   0.00     1.00          0.14          NA    0.26
                                                                (0.00) (0.00) (0.00)   (0.00)        (0.00)          NA  (0.00)

```

If I remove the <.NA.> it will work alright.

What is better behavior? Treat NA as a value when creating sets for diagnosis?

### LINKED ISSUE #304: term
I think we have problems with the current "term" approach for estimates taken from a model

 * First the estimands get odd names (`(Intercept)`) 
 * Second the estimands are  a pain to name  as you need to consult with model output; and the names are and not in the spirit of estimands based on potential outcomes
 * Third it seems to run into issues when combined with other appraoches as shown below
 * Fourth we don't need it. We are already able to do the same thing in a simpler way, see below

Is there anything against the simpler approach below?

```
# The issue:

population  <- declare_population(N = 100, X = rnorm(N))
potentials  <- declare_potential_outcomes(Y ~ (.25 + X) * Z + rnorm(N))
assignment  <- declare_assignment(m = 50)
design_stub <- population +  potentials + assignment + declare_reveal()

estimands_regression <- declare_estimand(
  `(Intercept)` = mean(Y_Z_0),
  `Z` = mean(Y_Z_1 - Y_Z_0),
  term = TRUE,
  label="Regression_Estimands")

estimators_regression <- declare_estimator(
  Y ~ Z,
  estimand = estimands_regression,
  model = lm,
  term = TRUE
)

estimand_2  <- declare_estimand(ATE = mean(Y_Z_1 - Y_Z_0),   label = "ate_simple")
estimator_2 <- declare_estimator(Y~Z, estimand = estimand_2, label = "dim")

design <- design_stub + estimands_regression + estimators_regression  + estimand_2 + estimator_2

# This has four rows instead of 3
simulate_design(design, sims = 1)

# This produces a warning and fails to connect estimates to estimand for dim
diagnose_design(design, sims = 2)

# Note that constituent parts work ok when declared separately
design_A <- design_stub + estimand_2 + estimator_2
diagnose_design(design_A, sims = 2)

design_B <- design_stub + estimands_regression + estimators_regression
diagnose_design(design_B, sims = 2)
```

# A simpler, better approach?

* Getting rid of term from the estimands declaration seems a lot simpler and gives users control over names
* Then in estimates declaration give the estimands of interest. "term" is a bit of a funny term for term. 
* With this usage, the older term "coefficients"  seems more accurate

```

estimands_regression <- declare_estimand(
  Q1 = mean(Y_Z_0),
  Q2 = mean(Y_Z_1 - Y_Z_0),
  label="Regression_Estimands")

estimators_regression <- declare_estimator(
  Y ~ Z,
  estimand = c("Q1", "Q2"),
  model = lm,
  term = TRUE
)

design <- design_stub + estimands_regression + estimators_regression + estimand_2 + estimator_2

# Now fine
simulate_design(design, sims = 1)

# Now fine
diagnose_design(design, sims = 2)

# To boot: nice flexibility to access different estimands 
estimator_3 <- declare_estimator(Y~Z, estimand = c(estimand_2, "Q2"), label = "dim2")
diagnose_design(design + estimator_3,  sims = 20)
```

acoppock:
This closes #304.

Edit: also closes https://github.com/DeclareDesign/DeclareDesign/issues/301

I'm getting an error in the tests saying that DesignLibrary can't find DeclareDesign, which I don't understand.

This also fixes a weird thing that if you have an estimand that has not estimators, it used to not be in the dx output, but now it is.

This approach might have unintended consequences, but I don't intend them.

### DIFF
--- a/R/diagnose_design.R
+++ b/R/diagnose_design.R
@@ -220,6 +220,7 @@ calculate_diagnosands <- function(simulations_df, diagnosands, group_by_set) {
 
 calculate_diagnosands_single_design <- function(simulations_df, diagnosands, group_by_set) {
   group_by_list <- simulations_df[, group_by_set, drop = FALSE]
+  group_by_list[is.na(group_by_list)] <- "NA"
 
   labels_df <- split(group_by_list, group_by_list, drop = TRUE)
   labels_df <- lapply(labels_df, head, n = 1)
--- a/R/diagnose_design.R
+++ b/R/diagnose_design.R
@@ -221,14 +221,10 @@ calculate_diagnosands <- function(simulations_df, diagnosands, group_by_set) {
 calculate_diagnosands_single_design <- function(simulations_df, diagnosands, group_by_set) {
   group_by_list <- simulations_df[, group_by_set, drop = FALSE]
   
-  not_all_na_columns <- apply(group_by_list, 2, function(x) ! all(is.na(x)))
-  
-  group_by_list <- group_by_list[, not_all_na_columns, drop = FALSE]
-
-  labels_df <- split(group_by_list, group_by_list, drop = TRUE)
+  labels_df <- split(group_by_list, lapply(group_by_list, addNA), drop = TRUE)
   labels_df <- lapply(labels_df, head, n = 1)
 
-  diagnosands_df <- split(simulations_df, group_by_list, drop = TRUE)
+  diagnosands_df <- split(simulations_df, lapply(group_by_list, addNA), drop = TRUE)
   diagnosands_df <- lapply(diagnosands_df, FUN = function(x) {
     dg <- diagnosands(x)
     setNames(dg[[2]], dg[[1]])
@@ -244,11 +240,11 @@ calculate_diagnosands_single_design <- function(simulations_df, diagnosands, gro
 
 calculate_sims <- function(simulations_df, group_by_set) {
   group_by_list <- simulations_df[, group_by_set, drop = FALSE]
-
-  labels_df <- split(group_by_list, group_by_list, drop = TRUE)
+  
+  labels_df <- split(group_by_list, lapply(group_by_list, addNA), drop = TRUE)
   labels_df <- lapply(labels_df, head, n = 1)
 
-  n_sims_df <- split(simulations_df, group_by_list, drop = TRUE)
+  n_sims_df <- split(simulations_df, lapply(group_by_list, addNA), drop = TRUE)
   n_sims_df <- lapply(n_sims_df, FUN = function(x) {
     data.frame(n_sims = nrow(x))
   })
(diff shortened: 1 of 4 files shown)


tests/testthat/test-diagnose-design.R
```diff
@@ -198,3 +198,15 @@ test_that("default diagnosands work", {
   diag <- diagnose_design(sims, sims = 5, bootstrap_sims = FALSE)
 })
 
+
+test_that("with and without term",{
+  library(DeclareDesign)
+  library(DesignLibrary)
```
nfultz: Generally we don't want library inside a test, use the skip_if_not_installed to check, and then do DesignLibrary::simple_factorial_designer.

R/diagnose_design.R
```diff
@@ -220,6 +220,7 @@ calculate_diagnosands <- function(simulations_df, diagnosands, group_by_set) {
 
 calculate_diagnosands_single_design <- function(simulations_df, diagnosands, group_by_set) {
   group_by_list <- simulations_df[, group_by_set, drop = FALSE]
+  group_by_list[is.na(group_by_list)] <- "NA"
```
nfultz: Are we sure that we aren't accidentally changing numbers into text bc of this?

nfultz: For pretty printing you might want `""` instead of `"NA"`.

nfultz (dismissed):

tests/testthat/test-diagnose-design.R
acoppock: great, thank you

R/diagnose_design.R
acoppock: ooh, i'll check about numbers and text

macartan: add_grouping_variables could contain columns with numbers.

tests/testthat/test-diagnose-design.R
acoppock: done

acoppock:
it turns out that this:

`split(group_by_list, lapply(group_by_list, addNA), drop = TRUE)`

solves the problem without messing wtih types or dropping columns.

nfultz (approved):
lgtm

PR #315, opened 2018-08-21
travis up for libunits

graemeblair:
wasn't installing sf correctly


nfultz (dismissed):
lgtm - you also might hoist the install command out of the build matrix.

PR #317, opened 2018-08-21
fix #316

### LINKED ISSUE #316: expand.design factors
`N` appears to be a factor:

```
big_data_designer <- function(N) {
  
  pop <- declare_population(
    N = N,  # sample size will grow
    u = rnorm(N), # unobserved heterogeneity that affects outcome and assignment
    e = rnorm(N, sd = 4) # ideosyncratic error
    )
  
  # outcomes depend on "u" and "e" (but not Z)
  pos <- declare_potential_outcomes(Y ~ 0 * Z + u + e)
  
  # Assignment is a function of "u"
  assignment <- declare_assignment(
    handler = function(data) {
      within(data, {
        Z = rbinom(n = nrow(data), size = 1, prob = pnorm(u))
      })
    }
    
  )
  reveal <- declare_reveal()
  
  # true effect is zero
  estimand <- declare_estimand(ATE = 0) 
  
  # use difference-in-means to estimate effect
  estimator <- declare_estimator(Y ~ Z, estimand = estimand)
  
  # declare the design
  design <- pop + pos + assignment + reveal + estimand + estimator
  design
}


# all sample sizes from 10 to 5,000 (we don't need to go bigger)
designs <- expand_design(designer = big_data_designer, N = seq(100, 5000, by = 100))

# we want to know about bias and power
diags <- declare_diagnosands(select = c("bias", "power"))

# this actually runs the simulations
dx <- diagnose_design(designs, diagnosands = diags, sims = 100, bootstrap_sims = FALSE)

str(dx$simulations_df)
```

acoppock:
needed to do type-convert for numeric parameters


nfultz (approved):
LGTM

PR #321, opened 2018-08-22
add get_estimates_data

graemeblair:
- adds a function `get_estimates_data` that takes a design (in future designs) and data. it returns a df.
- additional options allow you to only run some of the estimators via start (which step in design to start at) and end


R/design_helper_functions.R
```diff
@@ -130,6 +130,28 @@ run_design_internal.design <- function(design, current_df = NULL, results = NULL
...
+  
+  estimators <- sapply(design_stub, 
+                       function(x) attr(x, "causal_type") == "estimator")
+  
+  estimators <- design_stub[estimators]
+                       
+  results_list <- list()
+  for (i in seq_along(estimators)) {
+    results_list[[i]] <- estimators[[i]](data)
+  }
```
nfultz: Rather than yet another execution path, you can set up an execution_st and delegate to run_design_internal.

nfultz: https://github.com/DeclareDesign/DeclareDesign/blob/master/R/design_helper_functions.R#L105

You just need to provide results, start and current_df.

This will also handle the (inenvitable) edge cases where people have DGP steps in between estimator steps.

graemeblair: My sense is that is a good feature to have, but distinct from this one. I'd like this function to take a design and data (and optionally a start and stop) and *just* get the estimates based on that data. If there are intermediate dgp steps, this wouldn't run them and some estimates would potentially break. I think that's ok. You can subset the set of steps to run via start and stop. What do others think?

nfultz: In that case, you may as well slice the design up front, and then call run_design_internal on that thing - the whole method reduces to two lines. But it would essentially be running a different design than what the user provided.   

eg

```
d <- Filter(function(x) attr(x, "causal_type") == "estimator", design[start:end])
run_design_internal(d, data)$estimates_df
```

graemeblair: `run_design_internal` doesn't have a data argument; am I missing something? I've mostly implemented this, but obv. won't run given that issue.

nfultz: For historical reasons it was called current_df https://github.com/DeclareDesign/DeclareDesign/blob/master/R/design_helper_functions.R#L85

graemeblair: apologies, I was looking at the wrong function. thanks.

graemeblair:
I think this is good to go?

PR #325, opened 2018-08-22
cran prep

graemeblair:
- added news
- will use this PR for finalizing R CMD CHECK details


DESCRIPTION
```diff
@@ -19,7 +19,7 @@ Depends:
     fabricatr (>= 0.5.0),
     estimatr (>= 0.10.0)
 Imports:
-    rlang, methods
+    rlang
```
nfultz, line 22: Did you ever figure out why methods was needed in the first place?

graemeblair: I think there was an old importFrom call that was dropped at some point. Got an error in some builds on CRAN from this in the last version (all linux), but didn't seem to bother other builds.

PR #339, opened 2018-10-08
Gb/rename get estimates estimands

### LINKED ISSUE #337: get_estimates_data() using get_estimates()
Would it be possible to have sthg like 

```
get_estimates <- function(design, data = NULL) {
  if(!is.null(data) return(get_estimates_data(design, data)
  ...
```

ie so users do not have to call a separate function when they have data but get the same results when they provide data to get_estimates

graemeblair:
closes #337 

note this also ends the deprecation of `declare_design`


NAMESPACE
```diff
@@ -26,11 +25,12 @@ export(delete_step)
 export(diagnose_design)
 export(diagnose_designs)
 export(draw_data)
+export(draw_estimands)
+export(draw_estimates)
 export(expand_design)
 export(get_diagnosands)
 export(get_estimands)
```
nfultz, line 32: Is this stale, or did you add a "actuals" version for get_estimands?

R/design_helper_functions.R
```diff
@@ -144,7 +144,11 @@ run_design_internal.design <- function(design, current_df = NULL, results = NULL
 #'
 #' @rdname post_design
 #' @export
-get_estimates_data <- function(design, data, start = 1, end = length(design)) {
+get_estimates <- function(design, data = NULL, start = 1, end = length(design)) {
```
nfultz, line 147: ok, that makes sense I guess.

nfultz (approved):

PR #341, opened 2018-10-12
Minor doc updates

internaut:
You actually need to have installed both `future` *and* `future.apply` to run simulations in parallel. Just stumbled upon that while wondering why my simulations used only one CPU although I used `plan('multicore')`.

While trying to regenerate the docs, roxygen2 would throw an error. Fixed that by moving `@name` to end.


nfultz (dismissed):
LGTM - I'll clean up the merge conflicts and pull this in.

graemeblair:
thanks for this, @internaut!

graemeblair:
closing, fixed in #350

PR #345, opened 2018-10-17
posixct helper

### LINKED ISSUE #342: weird issue with posixct class
Was trying to use baseline data to do some simulations today and came across this really weird bug. My suspicion is that it's in the summary methods that are used to create the printout that happens once a design is declared using `+`. 

MWE:

```r
# Create dataset with dates
test <- data.frame(
  d = seq(from = as.POSIXct("2018-01-01"),
          to = as.POSIXct("2018-05-01"),
          by = "days"))
# Different ways of bringing data into DD
pop1 <- declare_population(data = test)
pop2 <- declare_population(handler = function() test) 
pop3 <- declare_population(handler = fabricate, data = test) 
# These all work great:
identical(test,fabricate(data = test))
identical(test,pop1())
identical(test,pop2())
identical(test,pop3())
# This breaks
pop + NULL
```
The error is the same one would get if one were running `round(test$d,digits = 2)`, which is why I think it must be that there is a `round()` call in the printed summary after the design is declared, which is choking on the posixct date class.

nfultz:
closes #342

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -96,6 +96,15 @@ describe_variable_impl.factor <- function(x, num_unique) {
   )
 }
 
+describe_variable_impl.POSIXct <- function(x, num_unique) {
+  data.frame(
+    as.list(as.character(summary(x))),
+    N_unique = num_unique,
+    check.names = FALSE
+  )
+}
+
+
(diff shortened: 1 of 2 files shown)


jaspercooper:
Great, thanks

PR #351, opened 2018-10-21
WIP add tab completion from DeclareDesign/completion

graemeblair:


graemeblair:
https://github.com/r-rudra/patch/blob/main/inst/embedded/usecases.R

PR #360, opened 2018-11-06
WIP generics v1

graemeblair:
- moving tidy to use generics package
- adding a fall-back tidy.default method that is used either if broom isn't loaded or in either case if there isn't a tidy.class method defined.


nfultz:
Instead of putting fallback code in tidy.default and exporting, I would strongly suggest keeping that function internal - you don't want to advertise/promise you can tidy arbitrary things. Instead, check with `hasMethod` if a method is available, and fallback outside of S3 dispatch.

nfultz:
See https://github.com/DeclareDesign/DeclareDesign/pull/361 for the general idea.

graemeblair:
so currently this doesn't work -- and we had gone the other direction -- because hasMethod("tidy", class(fit)) doesn't work for packages like `estimatr` that rely on generics. i.e. `tidy.lm_robust` doesn't appear in the list of methods. any suggestions?

nfultz:
ah, hasMethod is only for s4, sorry - I think you can write a small helper that works with S3 - eg

```
hasS3Method <- function(f, obj)
  for(i in class(obj)) {
    if(is.function(getS3method(f, i))) return(TRUE)
  }
  FALSE
}
```

graemeblair:
awesome, thanks

PR #372, opened 2018-12-11
Include lm_lin in validation_fn()

clarabicalho:
Prevent error when using lm_lin as estimator handler.

https://github.com/DeclareDesign/DesignLibrary/pull/214

### DIFF
--- a/R/declare_estimator.R
+++ b/R/declare_estimator.R
@@ -280,7 +280,7 @@ validation_fn(model_handler) <- function(ret, dots, label) {
 
     if (!quo_text(dots$model)
     %in%
-      c("lm", "glm", "lm_robust", "iv_robust", "difference_in_means", "horvitz_thompson")
+      c("lm", "glm", "lm_robust", "iv_robust", "difference_in_means", "horvitz_thompson", "lm_lin")
     &
       !requireNamespace("broom", quietly = TRUE)
     ) {
--- a/R/declare_estimator.R
+++ b/R/declare_estimator.R
@@ -273,18 +273,9 @@ validation_fn(model_handler) <- function(ret, dots, label) {
     model <- eval_tidy(dots$model)
     if (!is.function(model) || !"data" %in% names(formals(model))) {
       declare_time_error(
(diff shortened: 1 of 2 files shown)


nfultz:
I think lm_lin returns an class lm_robust object so this won't fix the issue? Not sure. @lukesonnet

clarabicalho:
Good point. That's why lm_lin works with tidy() without a special method, right? But I thought the issue with DesignLibrary here seems to be the input model --- why is lm_lin set to error with `validation_fn(model_handler)`? 

ps. DesignLibrary passes checks locally with this dd branch.

nfultz:
That makes sense. Can you add a test to DeclareDesign as well?

lukesonnet:
Yes `lm_lin()` returns an object of class `lm_robust`.

And I swear this is not passive aggressive, it is _only_ because I'm proud of myself for having documented it correctly, but the docs for `lm_lin()` confirms that this is what it returns.

graemeblair:
Thanks for this @clabima sorry for the long delay. We decided it was best to remove this error altogether, because it doesn't work in general. Now, there is an informative error pointing to broom at data time rather than declare time.

R/declare_estimator.R
```diff
@@ -369,10 +360,18 @@ hasS3Method <- function(f, obj) {
 # called by model_handler, resets columns names !!!
 fit2tidy <- function(fit, term = FALSE) {
   
+  # browser()
   if (hasS3Method("tidy", fit)) {
     tidy_df <- tidy(fit, conf.int = TRUE)
   } else {
-    tidy_df <- tidy_default(fit, conf.int = TRUE)  
+    tidy_df <- try(tidy_default(fit, conf.int = TRUE), silent = TRUE)
```
nfultz, line 367: This seems fine, but will swallow the original error message from tidy_default. You may want to also print it out to make future issues easier to triage.

nfultz (approved):
LGTM

graemeblair:
Thanks @nfultz!

PR #383, opened 2019-01-19
add draw_assignment, draw_sample, and expand draw_data functions

### LINKED ISSUE #347: draw_assignment and draw_sample
two functions similar to draw_estimates_data that 

1) draw_assignment: take data and then add assignment variable(s) (and assignment probs), running all assignment steps
2) draw_sample: take data and subset to sample (with inclusion probs), potentially more than once

graemeblair:
This PR closes https://github.com/DeclareDesign/DeclareDesign/issues/347

Adds:
- `draw_sample` which takes data and samples either all sampling steps (i.e. it subsets) by default, or just some of them using start and end step arguments;
- `draw_assignment` which takes data and assigns either all assignment steps (by default), or just some of them based on start and end step arguments;
- `draw_data` which takes data and conducts further data steps (either all of them by default, or a subset of them). Note that a design can technically have just a bunch of sampling and assignment steps and need not have a population or data at the top (given the way + works; we tested this), so it does make sense for you to be able to go from 1 to the end of the design starting with data you give. **`draw_data` by default runs the whole design, so now does or does not take data.**

Moved files and documentation around to more clearly document the `get_*` and `draw_*` steps.


graemeblair:
@macartan do you mind taking a look?

R/get_functions.R
```diff
@@ -0,0 +1,85 @@
...
+  
+  if(end < 1 || end > length(design)){
+    stop("Please provide an end step as a number between 1 and the total number of steps in the design.")
+  }
+  
+  design_subset <- Filter(function(x) attr(x, "causal_type") == "dgp", design[start:end])
+  
+  run_design_internal.design(design_subset, current_df = data, results = list(current_df = 0))$current_df
+  
+}
```
nfultz: These are all very similar, pull out the common logic to a helper function.

```
get_data_imp <- function(design, data = NULL, start = 1, end = length(design), pred) {

   if(is.null(data)){
    stop("Please provide a data frame to the data argument.")
  }

   if(start < 1 || start > length(design)){
    stop("Please provide a starting step as a number between 1 and the total number of steps in the design.")
  }

   if(end < 1 || end > length(design)){
    stop("Please provide an end step as a number between 1 and the total number of steps in the design.")
  }

   design_subset <- Filter(pred, design[start:end])

   run_design_internal.design(design_subset, current_df = data, results = list(current_df = 0))$current_df

 }

get_data <- function(design, data = NULL, start = 1, end = length(design)) {
  get_data_impl(design, data = NULL, start = 1, end = length(design),
    function(x) attr(x, "causal_type") == "dgp")
}
```

nfultz (commented):
Need to tidy up some of the duplicated logic.

R/get_functions.R
nfultz: For the predicates, using `%in%` instead of `==` is more defensive.

R/construct_design.R
```diff
@@ -21,6 +21,8 @@
 #'
 #' @return a list of two functions, the \code{design_function} and the \code{data_function}. The \code{design_function} runs the design once, i.e. draws the data and calculates any estimates and estimands defined in \code{...}, returned separately as two \code{data.frame}'s. The \code{data_function} runs the design once also, but only returns the final data.
 #'
+#' @name declare_design
```
nfultz, line 24: Is declare_design come back from the grave?

R/design_print_summary.R
```diff
@@ -0,0 +1,264 @@
+
```
nfultz, line 1: I usually try to keep reorganization to a seperate PR, otherwise the diffs get large and hard to follow, and history breaks.

R/construct_design.R
graemeblair: Hah, no just naming the file :)

R/design_print_summary.R
graemeblair: ah, thanks, tried to avoid and kept in a separate PR. will do moving forward.

graemeblair:
I think I implemented all this, but not understanding the sentinal value suggestion. 

`data %in% -9` works when you send -9 but when you send an actual data.frame it doesn't work given length:

```
In if (data %in% -9) { :
  the condition has length > 1 and only the first element will be used
```

Worry about false positives?

nfultz:
Whoops, should be `data ==` there

Edit:

or rather `identical()`

graemeblair:
ah got it thanks!

PR #394, opened 2019-02-02
add check to see if object sent is a design object

### LINKED ISSUE #393: Better error when a list of designs is sent to draw_data
This problem may also exist for other functions that can only work on a single design at a time.

graemeblair:
closes https://github.com/DeclareDesign/DeclareDesign/issues/393

### DIFF
--- a/R/get_functions.R
+++ b/R/get_functions.R
@@ -30,6 +30,7 @@
 #' @rdname get_functions
 #' @export
 get_estimates <- function(design, data = NULL, start = 1, end = length(design)) {
+  check_design_class_single(design)
   
   if(is.null(data)){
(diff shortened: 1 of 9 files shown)


R/get_functions.R
```diff
@@ -42,6 +43,7 @@ get_estimates <- function(design, data = NULL, start = 1, end = length(design))
 #' @rdname get_functions
 #' @export
 draw_assignment <- function(design, data = NULL, start = 1, end = length(design)) {
+  check_design_class_single(design)
```
nfultz: It's already in get_function_internal, so you don't need it in every single wrapper.

graemeblair: thx

PR #396, opened 2019-02-03
coerce estimator_label to character in reshape_diagnosis

### LINKED ISSUE #395: coerce estimator_label to factor?
Factors are creeping into diagnosis and causing problems including in `reshape_diagnosis` via `estimator_label` in custom estimators (and custom estimands), e.g.:

```
population <- declare_population(N = 100)

estimator <- declare_estimator(handler = function(data) {
  data.frame(estimator_label = "my_estimator", est = 5)
})

simulate_design(population + estimator, sims = 5) %>% as_tibble %>% head

# A tibble: 5 x 4
  design_label sim_ID estimator_label   est
  <chr>         <int> <fct>           <dbl>
1 design_1          1 my_estimator        5
2 design_1          2 my_estimator        5
3 design_1          3 my_estimator        5
4 design_1          4 my_estimator        5
5 design_1          5 my_estimator        5
```

Suggest we coerce with `as.character` at the `get_estimates/draw_estimates` stage. Objections?

Of course this can be fixed by the user via `stringsAsFactors = FALSE` but seems people forget this often.

graemeblair:
closes https://github.com/DeclareDesign/DeclareDesign/issues/395

### DIFF
--- a/R/get_functions.R
+++ b/R/get_functions.R
@@ -85,13 +85,13 @@ get_function_internal <- function(design, data = NULL, start, end, pred, results
   
   ret <- run_design_internal.design(design_subset, current_df = data, results = results)[[what]]
   
-  if(what == "estimates_df" && class(ret$estimator_label) != "character"){
+  if(what == "estimates_df" && !is.null(ret$estimator_label) && typeof(ret$estimator_label) != "character"){
     warning("The estimator label should be a character, but it is a ", 
             class(ret$estimator_label), 
             ". Try using handler = tidy_estimator(your_estimator_function)", call. = FALSE)
   }
   
-  if(what == "estimands_df" && class(ret$estimand_label) != "character"){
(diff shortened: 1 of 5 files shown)


R/get_functions.R
```diff
@@ -83,7 +83,20 @@ get_function_internal <- function(design, data = NULL, start, end, pred, results
   
   design_subset <- Filter(pred, design[start:end])
   
-  run_design_internal.design(design_subset, current_df = data, results = results)[[what]]
+  ret <- run_design_internal.design(design_subset, current_df = data, results = results)[[what]]
   
-}
+  if(what == "estimates_df" && class(ret$estimator_label) != "character"){
```
nfultz: Labels are optional, so don't print the warning if null.

graemeblair: ah, thanks

nfultz: Also character isn't really a class, its either a type or mode, I forget which.

graemeblair: ah yes thanks will fix

PR #399, opened 2019-02-05
take .data too; closes #398

### LINKED ISSUE #398: allow dplyr functions to be handlers
I think the only trouble is that we require handlers to have an argument that is exactly `data` and not `.data`.  Seems like this would be an easy fix, but maybe I'm missing something?  Seems like this is somewhere in `curry_data`, but I'm not sure.  @nfultz, do you have thoughts on this? No problem if not, thanks in any case.

```
library(DeclareDesign)
library(dplyr)

design1 <- 
  declare_population(N = 5, X = rnorm(N)) +
  declare_step(Y = 4, handler = fabricate)

draw_data(design1)

# errors
design2 <- 
  declare_population(N = 5, X = rnorm(N)) +
  declare_step(Y = 4, handler = mutate)

draw_data(design2)

# works
design3 <- 
  declare_population(N = 5, X = rnorm(N)) +
  declare_step(Y = 4, handler = function(data, ...)mutate(data, ... = ...))

draw_data(design3)
```

nfultz:


graemeblair (approved):
woah. awesome. LGTM! thx @nfultz

PR #405, opened 2019-02-28
WIP compare diagnoses

graemeblair:
opening PR for code review


R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+#' @export
+compare_diagnoses <- function(base_design,
+                              comparison_design, 
+                              sims = 1000,
+                              bootstrap_sims = 1000, 
+                              alpha = 0.01,
+                              merge_by_estimator = TRUE){
+  
+  
+  if(bootstrap_sims <= 99) stop("Please choose a higher number of bootstrap simulations")
```
graemeblair: In `diagnose_design` this is a warning and uses a different number. Should they be consistent?

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+ # Diagnose designs if base/comparison are of class "design"
+  
+  if( "design" %in% class(base_design)  ){
+   
+    diagnosis1 <- diagnose_design(base_design, 
+                                  sims = sims, 
+                                  bootstrap_sims = bootstrap_sims)
+
+    } else if(class(base_design) == "diagnosis") {
```
graemeblair: can you reindent all the code? a bit hard to follow. CMD+SHIFT+A in RStudio on the whole file will do the trick.

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+      if(base_design$bootstrap_sims<= 99)
+        stop("base_design must have a higher number of bootstrap simulations")
```
graemeblair: error should report that it must be above 99

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+    diagnosis1 <- base_design} else{
```
graemeblair: clean up spacing and line breaks

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+        stop("base_design must have a higher number of bootstrap simulations")
+    diagnosis1 <- base_design} else{ 
+    stop("base_design must be either a design or a diagnosis")}
+  
+
+  if( "design" %in% class(comparison_design)  ){
+    
+    diagnosis2 <- diagnose_design(comparison_design,
+                                  sims = sims, 
+                                  bootstrap_sims = 100)
```
graemeblair: why is 100 hard-coded here? should be an option

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+#' design_b <- replace_step(design_a, step = "assignment", declare_assignment(prob = 0.3) )
+#' 
+#' compare_diagnoses(design_a, design_b)
+#'  
+#' @export
+compare_diagnoses <- function(base_design,
+                              comparison_design, 
+                              sims = 1000,
+                              bootstrap_sims = 1000, 
+                              alpha = 0.01,
```
graemeblair: why .01 not .05?

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+#' 
+#' design_b <- replace_step(design_a, step = "assignment", declare_assignment(prob = 0.3) )
+#' 
+#' compare_diagnoses(design_a, design_b)
+#'  
+#' @export
+compare_diagnoses <- function(base_design,
+                              comparison_design, 
+                              sims = 1000,
+                              bootstrap_sims = 1000,
```
graemeblair: can these defaults be the same as `diagnose_design` to reduce confusion?

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  if( "design" %in% class(comparison_design)  ){
+    
+    diagnosis2 <- diagnose_design(comparison_design,
+                                  sims = sims, 
+                                  bootstrap_sims = 100)
+    
+    } else if(class(comparison_design) == "diagnosis"){
+      diagnosis2 <- comparison_design
+      } else{ 
+        stop("comparison_design must be either a design or a diagnosis")}
```
graemeblair: spacing

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+
+
+
+
+#' @export
+compare_diagnoses_internal <- function(diagnosis1, diagnosis2, merge_by_estimator, alpha) {
+  
+  # 1.Housekeeping
+  if(is.null(diagnosis1$bootstrap_replicates ))
+    stop("Can't compare diagnoses witouth bootstrap replicates")
```
graemeblair: this should be a more informative error, giving help for how to include bootstrap, i.e. "please set bootstrap_sims to a number above zero". also spelling.

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  if(length(diagnosis1$parameters_d[, "design_label"])  + length(diagnosis2$parameters_d[, "design_label"])> 2)
```
graemeblair: will this work in general if there is not a parameters_df?

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  # 1.Housekeeping
+  if(is.null(diagnosis1$bootstrap_replicates ))
+    stop("Can't compare diagnoses witouth bootstrap replicates")
+  
+  if(length(diagnosis1$parameters_d[, "design_label"])  + length(diagnosis2$parameters_d[, "design_label"])> 2) 
+    stop("Please only send design or diagnosis objects with one unique design_label.")
+
+  
+  diagnosands  <- base::intersect(diagnosis1$diagnosand_names,
```
graemeblair: call this diagnosand_names

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  # merge_by_set, used to merge diagnosands_df, must at least contain estimand_label
+  # at its largest possible cardinality, merge_by_set contains c("estimand_label", "term", " "estimator_label") 
+  # if group_by_set is different in both designs, merge_by_set contains labels the intersection from both.
+  # NOTE: Need to test how comparison is done when only on diagnosis has term or estimator_label
+  
+  group_by_set1 <- diagnosis1$group_by_set
+  group_by_set2 <- diagnosis2$group_by_set
+  
+  mand  <-  "estimand_label"  %in% group_by_set1 & "estimand_label"  %in% group_by_set2
```
graemeblair: please use full words here, following code style in rest of package

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  mator <-  "estimator_label" %in% group_by_set1 & "estimator_label" %in% group_by_set2
+  if( mand + mator < 1)
```
graemeblair: better would be to check what group_by_set1 and group_by_set2 have in common and then check to see if it includes estimand label and estimator label or both

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  # if group_by_set is different in both designs, merge_by_set contains labels the intersection from both.
+  # NOTE: Need to test how comparison is done when only on diagnosis has term or estimator_label
+  
+  group_by_set1 <- diagnosis1$group_by_set
+  group_by_set2 <- diagnosis2$group_by_set
+  
+  mand  <-  "estimand_label"  %in% group_by_set1 & "estimand_label"  %in% group_by_set2
+  mator <-  "estimator_label" %in% group_by_set1 & "estimator_label" %in% group_by_set2
+  if( mand + mator < 1) 
+    stop(paste0("diagnosands_df must contain at least estimand_label or estimator_label in common" ))
```
graemeblair: don't need paste0. clean up text.

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  merge_by_set <- NULL
+
+  if(mand) merge_by_set <- c("estimand_label")
```
graemeblair: please use curly braces for if's and no need for c()

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  if(merge_by_estimator){
+    if(!mator)
+      warning("Estimator label not used in columns for merging.
```
graemeblair: one line

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  mand  <-  "estimand_label"  %in% group_by_set1 & "estimand_label"  %in% group_by_set2
+  mator <-  "estimator_label" %in% group_by_set1 & "estimator_label" %in% group_by_set2
+  if( mand + mator < 1) 
+    stop(paste0("diagnosands_df must contain at least estimand_label or estimator_label in common" ))
+  
+  merge_by_set <- NULL
+
+  if(mand) merge_by_set <- c("estimand_label")
+  if(merge_by_estimator){
+    if(!mator)
```
graemeblair: mator == FALSE

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+      warning("Estimator label not used in columns for merging.
+              At least one of the designs doesn't contain estimator_label.")
+    else
```
graemeblair: curly braces

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+   merge_by_set <- c(merge_by_set, "estimator_label")
+  }
+  
+  if("term" %in% group_by_set1 & "term" %in% group_by_set2)
```
graemeblair: curly braces; clean up spacing

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  if("term" %in% group_by_set1 & "term" %in% group_by_set2) 
+    merge_by_set <- c(merge_by_set, "term")
+
+   
+  comparison_df <-  merge(diagnosis1$diagnosands_df  , diagnosis2$diagnosands_df, 
+                          by = merge_by_set, suffixes = c("_1", "_2"), stringsAsFactors = FALSE)
+  
+  if(nrow(comparison_df) == 0 ){
+    warning("Can't merge diagnosands data frames. Diagnoses don't have labels in common")
+   return(list(diagnosis1 = diagnosis1 , diagnosis2 = diagnosis2))
```
graemeblair: should just error, not return diagnoses -- the user already has access to both diagnoses -- and output should be predictable

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  }
+    
+  c_names  <- colnames(comparison_df)
+  suffix   <- c("_1", "_2")
+  dropcols <- grepl("[[:digit:]]+$", c_names ) |   c_names  %in% c("design_label", "estimand_label", "estimator_label", "term")
```
graemeblair: is there a cleaner way to do this? seems fragile

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+    
+  c_names  <- colnames(comparison_df)
+  suffix   <- c("_1", "_2")
+  dropcols <- grepl("[[:digit:]]+$", c_names ) |   c_names  %in% c("design_label", "estimand_label", "estimator_label", "term") 
+  
+  # Grab standard.error columns before droping cols that are not present in both diagnosands_df
+  # just in case diagnosis2 doesn't  include bootrstrap_replicates
+  filter_se      <- grepl("^se",  c_names)
+  diagnosands_se <-  comparison_df[, filter_se]
+  if(diagnosis2$bootstrap_sims == 0){
```
graemeblair: spacing

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  diagnosands_se1 <- diagnosands_se; diagnosands_se2 <- matrix(data = rep(NA, length(diagnosands_se)))
```
graemeblair: please line break and don't use semicolons

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  # Compute bootstrap confidence interval
+ 
+  bootstrap_df  <- diagnosis1$bootstrap_replicates
+  group_by_list <- bootstrap_df[,  group_by_set1, drop = FALSE]
+  labels_df     <- split(group_by_list, lapply(group_by_list, addNA), drop = TRUE)
+  labels_df     <- lapply(labels_df, head, n = 1)
+  bootstrap_df  <- split(bootstrap_df , lapply(group_by_list, addNA), drop = TRUE)
+  group_by_set  <- diagnosis1$group_by_set
+  lower.bound   <- lapply(bootstrap_df,FUN = function(x) {
```
graemeblair: lower.bound -> lower_bound

lower_bound and upper_bound should be the same function but use an argument for which you should return. they are identical except the quantile.

please do not use objects from the global environment within the function, include them as additional arguments to lapply (such as alpha and group_by_set)

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  lower.bound <- rbind_disjoint(lower.bound)
+  upper.bound <- rbind_disjoint(upper.bound)
+  
+  
+  # Split columns by diagnosands
+  filter_mean      <- gsub("_[[:digit:]]+$","", colnames(comparison_df)) %in%  diagnosands
+  diagnosands_mean <- comparison_df[,  filter_mean]
+  filter_sims      <- grepl("^n_sims", colnames(comparison_df))
+  sims_df          <- comparison_df[,filter_sims]
```
graemeblair: drop = FALSE?

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+   data.frame(comparison_df[pair, ], 
+              diagnosand = diagnosands,
+              t(mapply(function(mean_1, mean_2, se_1, se_2, l, u, n1 , n2) 
+                c(mean_1 = mean_1, mean_2 = mean_2, se_1 = se_1, se_2 = se_2,
+                  n_sims1 = n1, n_sims2 = n2, conf.low_1 = l, conf.upper_1 =  u, 
+                  in_interval =  ((mean_2 >= l) & (mean_2 <= u))), 
+                  #args
+                  mean_1 = diagnosands_mean[pair, 1:length(diagnosands)],
+                  mean_2 = diagnosands_mean[pair, (length(diagnosands)+1):ncol(diagnosands_mean)],
+                  se_1 = diagnosands_se1[pair,],
```
graemeblair: drop = FALSE here and throughout?

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+                  u =  upper.bound[pair, diagnosands],
+                  n1 = sims_df[pair, 1],
+                  n2 = sims_df[pair, 2])),
+                
+               stringsAsFactors = FALSE)})))
+  
+  c_names <- colnames(out)
+  
+  #comparison_df <-  rapply(object = out , f = round, classes = "numeric", how = "replace", digits = 2) 
+  comparison_df <- out
```
graemeblair: would prefer not to be renaming objects, can we keep it as a single name?

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  # Prepare output
+  # Add "_1" or "_2" to labels as needed.
+  if("estimand_label" %in% c_names & ! "estimand_label" %in% merge_by_set)
```
graemeblair: curly braces; spacing

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+  if( all(is.na(comparison_df$se_2))) comparison_df$se_2 <- NULL
+
+  attr(comparison_df, "alpha") <- alpha
+  
+  out <- list(compared.diagnoses_df =comparison_df ,
+              diagnosis1 = diagnosis1,
+              diagnosis2 = diagnosis2)
+
+  class(out) <- "compared.diagnoses"
```
graemeblair: diagnosis_comparison

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+
+#' @export
+print.summary.compared_diagnoses <- function(x, ...){
+  
+  bootstrap_rep1 <- x$diagnosis1$bootstrap_sims
+  bootstrap_rep2 <- x$diagnosis2$bootstrap_sims
+  n_sims1 <- nrow(x$diagnosis1$simulations_df)
+  n_sims2 <- nrow(x$diagnosis2$simulations_df)
+  x  <- x[["compared.diagnoses_df"]]
+  sx <- subset(x,  x[,"in_interval"] == 0)
```
graemeblair: please use more informative names

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  if(nrow(sx) == 0) {
+    print(paste("No divergences found ", atext)) 
+    return(x)}
+  
+  
+  if(n_sims1 == n_sims2 )
+    cat(paste0("\n Comparison of research designs diagnoses based on ", n_sims1, " simulations", atext))
+  else 
+    cat(paste0("\n Comparison of research designs diagnoses based on ", n_sims1, " simulations from `base_design` and ", n_sims2, " from `comparison_design`", atext))
+  if(bootstrap_rep1 == bootstrap_rep2)
```
graemeblair: curly braces throughout here

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  
+ # if(bootstrap_rep2 !=0) se_2 <- paste0("(",format_num(sx$se_2, 2), ")")
+  # else se_2 <- rep(length(sx$se_1), "-")
+  
+
+  sx   <- sx[ ,!colnames(sx) %in% c("in_interval", "se_1", "se_2", "n_sims1", "n_sims2", "conf.low_1", "conf.upper_1")]
+
+  
+  # Prep vars to shape data frame
+  mand  <- startsWith(colnames(sx), "estimand")
```
graemeblair: please use longer more informative names

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  out <- data.frame(rbind(sx, sx), stringsAsFactors = FALSE)
+  diagnosands_cols <- startsWith(colnames(out), "mean")
+  colnames(out)[diagnosands_cols] <- c("base", "comparison")
+  m <- nrow(out)
+  odd  <- (1:m) %% 2 != 0
+  even <- (1:m) %% 2 == 0
+  diagnosands_m <- sapply(sx[,diagnosands_cols], format_num, digits=2)
+  
+  
+  out[odd, diagnosands_cols ]  <- diagnosands_m
```
graemeblair: please follow reshape_diagnosis behavior here, this is a whole new set of behaviors and would be good to use the same to minimize code maintenance going forward

R/compare_diagnoses.R
```diff
@@ -0,0 +1,345 @@
...
+  # Force new factor 
+
+  out[,  designs] <- sapply(out[, designs] ,
+                            function(x) factor(x, levels = c(levels(x), " ")))
+  out[even,  designs] <-  " "
+    
+    
+  cat("\n\n")
+  print(out, row.names = FALSE)
+  cat(paste0("\n\n Displaying diagnosands that statistically diverge between base and comparison designs. See help file for details." ))
```
graemeblair: Shouldn't send to help file, should communicate the test used.

graemeblair (commented):
Thanks for this great work!

Lots of small style comments throughout, please excuse brevity!

Bigger suggestions:
1) add difference and standard error of difference via sqrt( se_1^2 + se_2^2 )
2) by default bootstrap both if designs are sent, so that the se of diff can be calculated

R/compare_diagnoses.R
nfultz: > CMD+SHIFT+A in RStudio on the whole file will do the trick.

Please do not do this - try to follow either the google style or hadley's. Or use Yihui's package - https://yihui.name/formatr/

R/compare_diagnoses.R
lilymedina: The function - as it currently is - checks which of the diagnosands of the`comparison_design` lie outside the (1-alpha)% bootstrap confidence interval of their counterparts in the `base_design` . So we do need bootstrap replicates (at least for the first design), and a high number of those (and sims) to gain precision and make compare_diagnoses stable.

lilymedina:
Thanks for reviewing. All small comments have been addressed. 

Re difference sounds good. Although I'm just not sure if it suggests a test of difference in means that we're not using.--I'll add columns anyway. 

**TODO:**
- [x] write tests
- [ ] add difference and se of difference columns

R/compare_diagnoses.R
graemeblair: Makes sense. Can the default # be the same? Right now it's 30 in `simulate_design` and 99 here.

R/compare_diagnoses.R
```diff
@@ -0,0 +1,331 @@
...
+                              alpha = 0.01,
+                              merge_by_estimator = TRUE){
+  
+  
+  if(bootstrap_sims_base <= 99) stop("Please choose a higher number of bootstrap simulations")
+  
+  
+ # Diagnose designs if base/comparison are of class "design"
+  
+  if( "design" %in% class(base_design)  ){
```
graemeblair: Please use check_design_class_single(base_design) here

R/compare_diagnoses.R
```diff
@@ -0,0 +1,331 @@
...
+    diagnosis1 <- diagnose_design(base_design, 
+                                  sims = sims, 
+                                  bootstrap_sims =  bootstrap_sims_base)
+
+    } else if(class(base_design) == "diagnosis") {
+      if(base_design$bootstrap_sims<= 99) stop("base_design must have a higher number of bootstrap simulations")
+    diagnosis1 <- base_design} else stop("base_design must be either a design or a diagnosis")
+  
+
+  if( "design" %in% class(comparison_design)  ){
```
graemeblair: check_design_class_single

R/compare_diagnoses.R
```diff
@@ -0,0 +1,331 @@
...
+  
+  if( "design" %in% class(base_design)  ){
+   
+    diagnosis1 <- diagnose_design(base_design, 
+                                  sims = sims, 
+                                  bootstrap_sims =  bootstrap_sims_base)
+
+    } else if(class(base_design) == "diagnosis") {
+      if(base_design$bootstrap_sims<= 99) stop("base_design must have a higher number of bootstrap simulations")
+    diagnosis1 <- base_design} else stop("base_design must be either a design or a diagnosis")
```
graemeblair: please add line breaks around the } -- very hard to read

R/compare_diagnoses.R
```diff
@@ -0,0 +1,331 @@
...
+    
+    diagnosis2 <- diagnose_design(comparison_design,
+                                  sims = sims, 
+                                  bootstrap_sims = bootstrap_sims_comparison)
+    
+    } else if(class(comparison_design) == "diagnosis")
+      diagnosis2 <- comparison_design else stop("comparison_design must be either a design or a diagnosis")
+  
+  
+  out <- compare_diagnoses_internal(diagnosis1, diagnosis2, merge_by_estimator, alpha )
```
graemeblair: please fix indenting in this area

R/compare_diagnoses.R
```diff
@@ -0,0 +1,331 @@
...
+#' @export
+#' @export
+compare_diagnoses_internal <- function(diagnosis1, diagnosis2, merge_by_estimator, alpha) {
+  # 1.Housekeeping
+  if (is.null(diagnosis1$bootstrap_replicates)) stop("Can't compare diagnoses witouth bootstrap replicates")
+  
+  if (length(diagnosis1$parameters_d[, "design_label"]) + length(diagnosis2$parameters_d[, "design_label"]) > 2) stop("Please only send design or diagnosis objects with one unique design_label.")
+  
+  
+  diagnosands <- base::intersect(diagnosis1$diagnosand_names, diagnosis2$diagnosand_names)
```
graemeblair: is base:: necessary here?

graemeblair:
Just checking in on this. A couple of things still need to be addressed, including:
1) cleaning up spacing and indentation to make the code more readable;
2) using curly braces around all ifs; 
3) simplifying lower.bound and upper.bound to a single function that is used twice (the code is identical except one number);
4) not exporting the internal function.

Let us know when you'd like us to comment again! Added a few but stopped because didn't want to interrupt the work in progress.

R/compare_diagnoses.R
lilymedina: Ignoring this suggestion since function should also work for objects of classe `diagnosis`

R/compare_diagnoses.R
lilymedina: so sorry!
Hope is ok now.

R/compare_diagnoses.R
nfultz: Also you can use the `inherits` function instead.

lilymedina:
Thanks for the review. Hoping code is easier to read now. 

Main updates are tests and mean and se difference columns; se difference is computed and returned if the `comparison_design` has bootstrap sims -- I still think these columns can be misleading as they might make people think that we are doing a mean comparison test.


Feel free to review and edit functions as needed so that it can be integrate with compare_designs.

acoppock:
Hi all -- we took a spin through this today prepping for a small DD update. Can I request 3 changes?

1. I would prefer it if the output did *not* subset to those rows where the diagnosands are "out of interval". This changes quite a bit from sim to sim, and in any case, I'd like to *see* that the diagnosands are similar, not just have them subset out.

2. Right now it appears that we show the estimate and standard error for the first design, but only the estimate for the second design.  Could we please show the estimate and standard error for both?

3. GB requested the difference and the standard error of the difference be added as a third column.  I think this is crucial, since that comparison is the main quantity of interest. I think that this should always be included. [I see a comment above that this could be toggled on and off, but I think it should always be displayed -- I mean, *aren't* we doing a pseudo mean comparison test with this confidence interval approach? Maybe I'm not following, sorry about that!]

Thank you!

PR #411, opened 2019-04-14
Fix params merge bug (eg when dplyr strips params attr)

### LINKED ISSUE #410: dplyr::mutate() causing a problem with diagnosing simulations dfs
Often you want to `mutate()` a simulations df before diagnosing (for example, simulation-level p-value adjustment). 

Something a bit weird happening when you use `dplyr::mutate()` on a simulations df and pass it to `diagnose_design()`. Piping into `data.frame()` or `as.data.frame()` doesn't appear to fix it.

MWE:

```r
library(tidyverse)
design <- 
  declare_population(
    N = 100, 
    X1 = rnorm(N),
    X2 = rnorm(N)
  ) +
  declare_estimand(true_effect = 0) +
  declare_estimator(X1 ~ X2, model = lm_robust, estimand = "true_effect", label = "X1 on X2") 

simulations <-   
  simulate_design(design, sims = 20) %>% 
  mutate(a = 1) 

diagnose_design(simulations)
```

Gives me: `Error in fix.by(by.y, y) : 'by' must specify a uniquely valid column`

nfultz:
Closes #410


nfultz:
@graemeblair @acoppock The travis failure is something with data.table on 3.4 only - this (and #412 ) should be safe to merge. Support for 3.4 ends when 3.6 is released in two weeks.

PR #412, opened 2019-04-14
diffobj should go in Suggests, not Imports

nfultz:
This makes the diffobj package a `Suggest` instead of `Import`


graemeblair:
Sounds good -- thanks @nfultz!

PR #416, opened 2019-05-07
Compare diagnoses2

lilymedina:
Alternative to PR [#405](https://github.com/DeclareDesign/DeclareDesign/pull/405)

this version of compare_diagnoses computes the difference in diagnosands between two bootstrap_replicates and returns the mean and se of the difference indicating whether the difference is significantly different from 0 (see output below).

Edits and comments welcome!
```
design_a <- two_arm_designer(N = 30)
design_b <- two_arm_designer(N = 40)
comparison <- compare_diagnoses(design_a, design_b)

comparison
```

```
 Comparison of research designs diagnoses based on 500 simulations
Diagnosand estimates with bootstrapped standard errors in parentheses (100).

 estimand_label estimator_label term    diagnosand design1 design2 difference
            ATE       estimator    Z          bias   -0.00   -0.00      -0.00
                                                    (0.02)  (0.01)     (0.02)
            ATE       estimator    Z      coverage    0.94    0.95       0.00
                                                    (0.01)  (0.01)     (0.01)
            ATE       estimator    Z mean_estimand    1.00    1.00       0.00
                                                    (0.00)  (0.00)     (0.00)
            ATE       estimator    Z mean_estimate    1.00    1.00      -0.00
                                                    (0.02)  (0.01)     (0.02)
            ATE       estimator    Z       mean_se    0.36    0.32   -0.05***
                                                    (0.00)  (0.00)     (0.00)
            ATE       estimator    Z         power    0.75    0.85    0.10***
                                                    (0.02)  (0.02)     (0.02)
            ATE       estimator    Z          rmse    0.37    0.32   -0.05***
                                                    (0.01)  (0.01)     (0.01)
            ATE       estimator    Z   sd_estimate    0.37    0.32   -0.05***
                                                    (0.01)  (0.01)     (0.01)
            ATE       estimator    Z   type_s_rate    0.00    0.00       0.00
                                                    (0.00)  (0.00)     (0.00)

Note: *p < 0.10, ** p < 0.05, *** p < 0.01



```


macartan:
this output looks really nice

is it right to call this a p value however?  the bootstrap distribution is
not a distribution under the null, but that's not the case here is it?

wonder if we should just flag these cases to users rather than claiming a p
value

acoppock:
Agreed, output is nice. Thank you for implementing this!

Can we move to just one star, corresponding to p <0.05? 

looks like the p-values are calculated as:
`p.value <- ifelse(sign(mu) == -1, mean(diff >= 0),  mean(diff <= 0))`

Which isn't quite a p-value. (I'm the one who asked for this calculation, now I'm not sure if it's what we want!). Should we just switch to a normal approximation and do

`p.value <- 2*pnorm(abs(estimate/std.error), lower.tail = FALSE)`

instead, so that it corresponds to what people are expecting?  I'm not sure what the interpretation of the original value should be.

macartan:
no reason though for a normal to hold here in general as CLT doesn't apply.

why not just ask if 0 is in the [.025, .0975] bootstrap confidence interval
and if not, flag it

lilymedina:
That's right, we can't always rely on the clt. So I'll compute the 95% ci and flag those cases containing zero.

Thanks for reviewing!

lilymedina:
oops true.

nfultz:
> Can we move to just one star, corresponding to p <0.05?

I would suggest matching base R as much as possible in this case. 

We should format the p-vals using scientific notation where appropriate using:

https://www.rdocumentation.org/packages/base/versions/3.6.0/topics/format.pval

Calculate the stars the same way:

		Signif <- symnum(pv, corr = FALSE, na = FALSE,
				 cutpoints = c(0,  .001,.01,.05, .1, 1),
				 symbols   =  c("***","**","*","."," "))

from https://github.com/wch/r-source/blob/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/library/stats/R/anova.R#L149-L152

And respect the same options: `options(show.signif.stars=FALSE)`

see also stats::printCoefmat

R/compare_diagnoses.R
```diff
@@ -0,0 +1,278 @@
...
+  sims <- nrow(x$diagnosis1$simulations_df)
+  compared_diagnoses_df <- x[["compared_diagnoses_df"]]
+
+  column_names <- colnames(compared_diagnoses_df)
+  identifiers <- column_names %in% c("estimand_label", "term") | grepl("^estimator_label", column_names)
+  identifiers_columns <- column_names[identifiers]
+  p.value <- compared_diagnoses_df$p.value
+  stars <- ifelse(p.value < 0.01 ,"***",
+                  ifelse(p.value < 0.05 & p.value>=0.01, "**",
+                         ifelse(p.value < 0.1 & p.value>=0.05, "*", "")))
```
nfultz: We should match the stars to base R / `stats::lm` - a p of .08 is not a `*`

See comment in discussion thread.

lilymedina:
This is how the output looks like with the latest committed changes.

```
Comparison of research designs diagnoses based on 500 simulations
Diagnosand estimates with bootstrapped standard errors in parentheses (100).

 estimand_label estimator_label term    diagnosand design1 design2 difference
            ATE       estimator    Z          bias   -0.02    0.01       0.03
                                                    (0.02)  (0.01)     (0.02)
            ATE       estimator    Z      coverage    0.95    0.96       0.01
                                                    (0.01)  (0.01)     (0.01)
            ATE       estimator    Z mean_estimand    1.00    1.00       0.00
                                                    (0.00)  (0.00)     (0.00)
            ATE       estimator    Z mean_estimate    0.98    1.01       0.03
                                                    (0.02)  (0.01)     (0.02)
            ATE       estimator    Z       mean_se    0.36    0.31     -0.05*
                                                    (0.00)  (0.00)     (0.00)
            ATE       estimator    Z         power    0.75    0.87      0.13*
                                                    (0.02)  (0.02)     (0.03)
            ATE       estimator    Z          rmse    0.37    0.30     -0.06*
                                                    (0.01)  (0.01)     (0.01)
            ATE       estimator    Z   sd_estimate    0.37    0.30     -0.06*
                                                    (0.01)  (0.01)     (0.01)
            ATE       estimator    Z   type_s_rate    0.00    0.00       0.00
                                                    (0.00)  (0.00)     (0.00)

Note: * indicates whether the 95% bootstrap confidence interval of the difference in diagnosands does not contain zero 
```

R/compare_diagnoses.R
```diff
@@ -0,0 +1,291 @@
...
+                 bootstrap_sims = bootstrap_sims,
+                stringsAsFactors = FALSE)
+     
+      }))
+    suppressWarnings(cbind(identifiers, difference))
+
+  }
+  
+  return_df <- lapply(names(comparison_list),  create_difference_df )
+  return_df <- data.frame(rbind_disjoint(return_df, infill = ""))
```
acoppock: just a question -- is this call to data.frame required?  I thought rbind_disjoint always returned a data.frame, but I could be wrong

acoppock:
Very nice -- thanks so much for this work.

Re: Neal's comment, we use `alpha` in estimatr, so I think that would be appropriate here too, no?

R/compare_diagnoses.R
nfultz: AFAICT rbind_disjoint always returns a data.frame.

lilymedina:
Great! I'll incorporate comments and write tests. Thanks!

tests/testthat/test-compare-diagnoses.R
```diff
@@ -0,0 +1,72 @@
...
+  expect_error(compare_diagnoses(design_a, design_c)) 
+  design_d <- delete_step(design_c, "estimator")
+  expect_warning(compare_diagnoses(design_c, design_d, sims = 40)) 
+})
+
+
+test_that("compare_diagnoses prints", {
+ 
+  print <- capture.output(DeclareDesign:::print.compared_diagnoses(comparison))
+  expect_is(print, "character")
```
nfultz: capture.output always returns a character, so this test can never fail as far as I can tell.

Instead, I would use expect_output and check something appears in the output (`Note: * indicates` or another constant would be ok).

PR #422, opened 2019-09-06
diff obj fixes

graemeblair:
- diffobj had some issues related to printing data frames, caught by the tests. this PR resolves them. (it was printing df's side-by-side but they look terrible that way and it complained).

### DIFF
--- a/R/compare_designs.R
+++ b/R/compare_designs.R
@@ -1,8 +1,51 @@
-compare_partial <- function(FUN, DIFFFUN){
-  function(design1, design2, format = "ansi256", mode = "sidebyside", pager = "off", context = -1L, rmd = FALSE){
-    stopifnot(requireNamespace("diffobj"))
-    DIFFFUN <- get(DIFFFUN, getNamespace("diffobj"))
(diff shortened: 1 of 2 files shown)


R/compare_designs.R
```diff
@@ -1,8 +1,51 @@
-compare_partial <- function(FUN, DIFFFUN){
-  function(design1, design2, format = "ansi256", mode = "sidebyside", pager = "off", context = -1L, rmd = FALSE){
-    stopifnot(requireNamespace("diffobj"))
-    DIFFFUN <- get(DIFFFUN, getNamespace("diffobj"))
-    compare_design_internal(FUN, DIFFFUN, design1, design2, format = "ansi256", mode = "sidebyside", pager = "off", context = -1L, rmd = FALSE)
+compare_partial <- function(FUN, DIFFFUN, is_data = FALSE){
```
nfultz, line 1: I would just make two functions and call the appropriate one - see also https://martinfowler.com/bliki/FlagArgument.html

nfultz (changes requested):

R/compare_designs.R
graemeblair: fair, will do

PR #426, opened 2019-10-19
add error when a simulations_df is sent with zero rows

graemeblair:
This happened when I did simulate_design() then diagnose_design(na.omit(sims)). Edge case, but it was a weird error when it happened.


nfultz (dismissed):
LGTM

nfultz (dismissed):
LGTM

PR #431, opened 2020-03-22
fixes CRAN errors due to stringsAsFactors changes

graemeblair:


nfultz:
Looks like there's a build error on oldrel - pbkrtest is not available - can you add a condition to the test that uses it?

graemeblair:
It doesn't look straightforward to do this: the error happens at install dependencies time, so doesn't get to tests. And we don't suggest it, it must be a subdependency. Welcome suggestions!

PR #435, opened 2020-04-29
post estimation

graemeblair:
We've implemented a small change to how declare_estimator works to enable users to use their own post-estimation functions, rather than being locked in to our fit2tidy() function that helped tidy the results of model fits. Instead, they can use our default function, which is essentially the same as before, or they can use a custom post-estimation command, or tidy or glance from the broom package directly. The biggest advantage of this is that in declare estimator you can use glance, which instead of summarizing coefficients provides model fit statistics, i.e.:

`declare_estimator(Y ~ Z + covariate, model = lm_robust, model_summary = glance)`

which returns:

```
  estimator_label r.squared adj.r.squared statistic      p.value df.residual   N se_type
1       estimator 0.1893212     0.1726062  11.07726 4.645896e-05          97 100     HC2
```


R/declare_estimator.R
```diff
@@ -264,29 +266,30 @@ model_handler <-
     
     # following copied from dplyr:::as_inlined_function and dplyr:::as_fun_list
     
-    if(rlang::is_formula(post_estimation)) {
+    if(is_formula(post_estimation)) {
```
nfultz: I'd recommend pulling the formula/function/character logic out into a helper function.

R/declare_estimator.R
```diff
@@ -244,8 +244,10 @@ tidy_estimator <- function(estimator_function) {
 
 #' @param data a data.frame
 #' @param model A model function, e.g. lm or glm. By default, the model is the \code{\link{difference_in_means}} function from the \link{estimatr} package.
+#' @param post_estimation A model-in data-out function to extract coefficient estimates or model summary statistics, such as \code{\link{tidy}} or \code{\link{glance}}. By default, the \code{DeclareDesign} post-estimation function \code{\link{tidy_filter}} is used, which tidies data and optionally filters to relevant coefficients.
```
nfultz: Add a note to the docs that term is threaded through to post_estimation, it can be hard to trace which function gets called with what.

nfultz (approved):

DeclareDesign.Rproj
```diff
@@ -16,3 +16,6 @@ BuildType: Package
 PackageUseDevtools: Yes
 PackageInstallArgs: --no-multiarch --with-keep.source
 PackageRoxygenize: rd,collate,namespace
+
+QuitChildProcessesOnExit: Yes
+DisableExecuteRprofile: Yes
```
nfultz, line 21: Where did these come from?

nfultz (dismissed):

nfultz:
One other thought, the phrase "post-estimation" is really a stata-ism, would meditate on if really want to call it that, as this is not exactly the same thing. I would recommend "tidyer" or similar. It might make the most sense to separate the term-filter at the end from the tidy function / fallback implementation, and just have the user provide their own tidy method if they so choose.

Other ideas / words would be extract, summary / summarise (although that's used elsewhere), distill, clean, tidy, convert,

graemeblair:
Thanks @nfultz this is super helpful

PR #436, opened 2020-04-29
add declare_test

graemeblair:
This function is an alias for declare_estimator to reflect the fact that often we're interested in conducting a test without a specific estimand in mind. https://twitter.com/jakewbowers/status/1249892904308482049?s=20


nfultz:
LGTM. Maybe add a little more to the roxygen, I think the intent is that this is for non-model-based nhst? It could use some guidance oe when to use this vs declare_estimator.

nfultz (dismissed):

PR #438, opened 2020-04-30
change to transparently use tidy_estimator and tidy_test in the creation of declare_test and declare_estimator

graemeblair:
- removes the intermediary function `estimator_handler` and directly uses `tidy_estimator(model_handler)` and `tidy_test(model_handler)` in the calls for `declare_estimator` and `declare_test` respectively. The virtue is that it is more clear to the user where things are going and signals they can use these tidy_* functions themselves.
- adds better docs to both declare functions to help users understand what's being passed around where.


R/declare_test.R
```diff
@@ -78,13 +80,64 @@
...
+#' \code{tidy_test} takes a data-in-data out function to \code{fn}, and returns a data-in-data-out function that first runs the provided test function \code{fn} and then appends a label for the test.
 #'
-#' @param fn A function that takes a data.frame as an argument and returns a data.frame with the test statistic (and other inferential statistics) and a label.
+#' @param fn A function that takes a data.frame as an argument and returns a data.frame with test statistics as columns.
 #' @rdname declare_test
 #' @export
-tidy_test <- tidy_estimator
+tidy_test <- function(fn) {
+  if (!("data" %in% names(formals(fn)))) {
+    stop("Must provide a `test_function` function with a data argument.")
```
nfultz, line 103: If the only difference is the error message, you can parameterize the string instead of duplicate the code.

R/declare_test.R
```diff
@@ -78,13 +80,64 @@
...
 #' }
 #' 
-declare_test <- declare_estimator
+#' @seealso See \code{\link{declare_estimator}} for documentation of the \code{model_handler} function.
+#' 
+declare_test <- 
+  make_declarations(
+    tidy_test(model_handler),
+    step_type = "estimator",
+    causal_type = "estimator",
```
nfultz, line 89: Depending how far you want to take this, you could make a new causal_type also, so that "tests" are aggregated and tagged differently from estimates. That would allow different sets diagnostics on each (eg no bias/coverage on tests).  Someone will for sure ask how to not include a bunch of NA columns for test-only designs.

R/declare_estimator.R
```diff
@@ -261,6 +261,18 @@ tidy_estimator <- function(fn) {
...
 
+#' @rdname ourPkg-deprecated
+#' @section \code{tidy_estimator}:
+#' For \code{tidy_estimator}, use \code{\link{label_estimator}}.
+#'
+#' @importFrom lifecycle deprecate_soft
+#'
+#' @export
+tidy_estimator <- function(estimator_function) {
+  deprecate_soft("1.0.0", "tidy_estimator()", "label_estimator()")
```
nfultz: Seems fine, but I'd recommend just using message/warn/stop/.Deprecated from base R instead of pulling in an extra tidyverse library just for a couple messages, especially if the intent is to remove them further down the road. Also this will cause the 1-2 designs in DesignLibrary that were using this to start throwing warnings, CRAN will probably flag that when you submit.

R/deprecated.R
```diff
@@ -1,24 +1,20 @@
...
-    old = "get_estimands",
-    msg = paste0(
-      "\nThe get_estimands function has been renamed draw_estimands, to make clear ",
-      "that the estimands are draws from a simulation. Please replace your code with draw_estimands."
-    )
-  )
-  
-  cat()
+get_estimands <- function(design) {
+  deprecate_stop("1.0.0", "get_estimands()", "draw_estimands()")
```
nfultz: This has been marked deprecated for a while, you can safely remove it at this point. I would add an alias to draw_estimands roxygen so that ?get_estimands resolves to the new function.

nfultz (commented):
So far so good.

R/declare_test.R
graemeblair: They're no longer the same so not doing this for now but will think about how to simplify for future

nfultz (dismissed):
LGTM

PR #442, opened 2020-07-13
declare_measurement

acoppock:
For a long time, we've been putting measured variables in declare_population. Of course that works just fine computationally. We now have a specific declare_measurement in order to draw the conceptual distinction between latent variables in the M that become data when a data strategy D is applied.

declare_measurement has a fabricate handler.  You can use this in many of the places where we've been doing a declare_step to introduce a measurement procedure in the declaration, plus it saves typing handler = fabricate


nfultz:
LGTM.  

How do you foresee using this in combination w/ potential outcomes? POs create the conditional distribution, measurement does the draw?

acoppock:
Thanks Neal.  

One way I think this might interact with POs goes like this: The pos Y_Z_0 and Y_Z_1 are of course latent.When we "declare_reveal" we switch which latent potential outcome eventuates -- but we still haven't measured it. we can use declare_measurement to make that step explicit. 

so you could go

declare_potential_outcomes() +
declare_assignment() +
declare_reveal()+ # this "reveals" an unmeasured Y in some scale
declare_measurement() # this takes Y and turns it into a Likert or something.


we don't have to do this every time of course and we might want to be defining estimands in terms of the outcomes as they will be measured, but it's a useful piece of technology to emphasize the measurement part of the design (which is what I was trying to do, which is how we ended up here!)

PR #455, opened 2020-11-23
Use a data mask for estimand intermediate values

### LINKED ISSUE #454: estimands should be a function of other estimands
```
mands <- declare_population(N = 5) + declare_estimands(estimand1 = rnorm(1), estimand2 = 5*estimand1)
draw_estimands(mands)
Error: Error in step 2 (estimand):
	Error in eval_tidy(options[[i]], data = data): object 'estimand1' not found
```

In fabricate(), later objects can be a function of earlier variables, seems like this should be possible here too. Applications include ratio of estimands.

nfultz:
Closes #454


graemeblair:
incredible. thank you @nfultz!

graemeblair:
the build is failing because of sf in latest; I will look into what broke and fix

PR #458, opened 2020-12-20
[WIP] suggested new syntax

acoppock:
proposed changes to fabricatr:

- add potential_outcomes
- add reveal_outcomes 
- allow "unnamed" objects in fabricate so that potential_outcomes can reveal multiple POs [causes problems in tests right now]

proposed changes to DeclareDesign:

- add declare_model (copy of declare_population with goal of removing use of declare_potential_outcomes and switch to a single function for population and POs with use of the new fabricatr::potential_outcomes)
- add declare_inquiry (copy of declare_estimand)
- change default assignment handler to fabricate
- change default sampling handler to fabricate
- change way outcomes are revealed usually to be in declare_measurement (using new fabricatr::reveal_outcomes)

```
des <- 
  declare_model(
    N = 100, 
    U = rnorm(N),
    potential_outcomes(Y ~ 0.1 * Z + U)
  ) + 
  declare_inquiry(ATE = mean(Y_Z_1 - Y_Z_0)) +
  declare_sampling(S = complete_rs(N = N, n = 80)) +
  declare_assignment(Z = complete_ra(N, m = 40)) +
  declare_measurement(Y = reveal_outcomes(Y ~ Z)) +
  declare_estimator(Y ~ Z, model = lm_robust)
```


nfultz:
I'm generally in favor of moving utilities into fabricatr. Here is  a couple notes:

* `declare_model` - model usually refers to eg linear models on the analysis side; this name doesn't emphasize that it is for DGP.

* `declare_sampling` - fabricatr doesn't do subsetting IIRC, so this will probably still need the custom handler that does that part of it.

* You will also need to update all the validation logic to match the new handler signatures. Consider removing them if they aren't actually helpful.

Overall, this is a pretty large change; I would recommend bumping the major version. To get this on to cran, you will need to push fabricatr, DeclareDesign *and* DesignLibrary.

R/declare_sampling.R
```diff
@@ -37,13 +37,32 @@
...
+
+  options <- quos(...)
+  
+  data <- fabricate(data = data, !!!options, ID_label = NA)
+  
+  rows <- enquo(subset)
+  rows_val <- eval_tidy(rows, data)
+  stopifnot(is.logical(rows_val))
+  
+  data[rows_val, , drop = FALSE]
```
nfultz, line 74: NAs will propogate here.

NAMESPACE
```diff
@@ -13,6 +13,7 @@ S3method(str,design_step)
 S3method(summary,compared_diagnoses)
 S3method(summary,design)
 S3method(summary,diagnosis)
+export(assignment_handler)
```
nfultz: handlers are generally internal to the package

R/diagnosis_helper_functions.R
```diff
@@ -127,6 +127,10 @@ reshape_diagnosis <- function(diagnosis, digits = 2, select = NULL, exclude = NU
     levels(return_df[[i]]) <- c(levels(return_df[[i]]), "")
   }
   return_df$estimator_label <- as.character(return_df$estimator_label)
+  
+  for(j in c(sort_by_list, parameter_names)) if(is.factor(return_df[[j]]) && !"" %in% levels(return_df[[j]]))
```
nfultz, line 129: parameter_names is already handled four lines above. Can unify the two loops.

PR #460, opened 2020-12-28
[WIP] test of wallet functionality

graemeblair:
This PR is a proof-of-concept of "wallet" functionality - steps that add objects to an environment. The main use case for now is declaring a statistical model. This PR enables this code (see last two lines):

```
library(DeclareDesign)

des <- 
  declare_population(
    N = 100, 
    U = rnorm(N)
  ) + 
  declare_potential_outcomes(Y ~ 0.1 * Z + U) + 
  declare_estimand(ATE = mean(Y_Z_1 - Y_Z_0)) +
  declare_sampling(S = complete_rs(N = N, n = 80)) +
  declare_assignment(Z = complete_ra(N, m = 40)) +
  declare_measurement(Y = reveal_outcomes(Y ~ Z)) +
  declare_estimator(Y ~ Z, model = lm_robust) + 
  declare_statistical_model(OLS = lm_robust(Y ~ Z)) + 
  declare_statistical_model_summary(tidy(OLS), subset = term == "Z")

head(draw_estimates(des))
```

This attach(wallet); detach("wallet") code seems to substantially slow things down. May need to find a faster way to do this, but have explored many dead ends.


R/declare_statistical_model.R
```diff
@@ -0,0 +1,27 @@
...
+declare_statistical_model <- make_declarations(statistical_model_handler, step_type = "wallet", causal_type = "estimator")
+
+#' @importFrom rlang eval_tidy quos as_label
+#' @export
+statistical_model_handler <- function(data, ...) {
+  
+  dots <- quos(...)
+  
+  if(length(dots) > 1) {
+    stop("Please only send one model to declare_statistical_model.")
```
nfultz, line 12: May as well just loop over the dots and evaluate all of them, instead of stopping here.

R/declare_statistical_model.R
```diff
@@ -0,0 +1,27 @@
...
+  expr_quo <- dots[[1]]
+  
+  # label estimator
+  nm <- names(dots)[1]
+  if(nm == "") {
+    nm <- as_label(expr_quo)
+  }
+  
+  mod <- list()
+  mod[[nm]] <- eval_tidy(expr_quo, data = data)
```
nfultz, line 24: can replace this whole block of logic with good use of `:=`

nfultz:
`attach`/`detach` are verboten on CRAN - I think to make this work, you can use overscoping when you evaluate the step.  The wallet should also be declared on the `results` vector and not a local variable, because the full set of steps may not all be executed in a single call, depending on the user's execution plan.

graemeblair:
Thanks @nfultz. Any hints on what we would do to use overscoping when eval'ing the next step? Tried a number of combinations of making data masks and creating environments linked to the wallet in various ways to no avail.

nfultz:
> Thanks @nfultz. Any hints on what we would do to use overscoping when eval'ing the next step? Tried a number of combinations of making data masks and creating environments linked to the wallet in various ways to no avail.

The next thing I would try is to overscope the quosure within the next_step environment.

graemeblair:
Hi @nfultz I've been playing around with this for the last few days with no luck. Did you mean something like this? 

```
step <- as_overscope(step, data = wallet)
current_df <- step(current_df)
```

nfultz:
Instead of modifying the step, you can control the evaluation. something along the lines of

    current_df <- eval_tidy(step(current_df), data=as_data_mask(wallet))

would be ideal, but I think standard scoping rules would make that not work.

If you look at the code for a step:

> step <- declare_population(N=20, x=1)
> body(step)
{
    eval_tidy(if (is.null(data) & is_implicit_data_arg(dots)) 
        quoNoData
    else quoData, data = list(data = data))
}

which is found here: https://github.com/DeclareDesign/DeclareDesign/blob/dd219609be6245c486d472879031230273bc76a8/R/aaa.R#L100

You can modify that line to have wallet overscope the quosure. Then you would need to add wallet to that functions arguments, and call it via

    current_df <- step(current_df, wallet)

PR #462, opened 2021-01-30
arguments for diagnose design

macartan:
`make_groups` passes a quoted expression to fabricate, and

`add_groups` is a renaming of add_grouping_variables which is a mouthful -- backwards compatibility issue here but it is a little used argument I think

`select` just passes a select list to reshape_diagnosis

result:
```
> diagnosis <- diagnose_design(design, 
+   make_groups = "significant = p.value <= 0.05",
+   add_groups = "significant",
+   select = "Bias"
+   )
> diagnosis

Research design diagnosis. Diagnosand estimates with bootstrapped standard errors in parentheses (100 replicates).

 Design Label Estimand Label Estimator Label Term Significant N Sims   Bias
       design            ATE       estimator    Z       FALSE    454  -0.01
                                                                     (0.00)
       design            ATE       estimator    Z        TRUE     46   0.10
                                                                     (0.01)
```

issues: not loving that makegroups takes a quoted expression; `make_groups(significant = p.value <= 0.05)` would be more natural


R/diagnose_design.R
```diff
@@ -94,18 +102,21 @@ diagnose_design <- function(...,
     simulations_df <- simulate_design(!!!dots, sims = sims)
     diagnosands <- setup_diagnosands(!!!dots, diagnosands = diagnosands)
   }
-
+  
   # figure out what to group by ---------------------------------------------
+  # Optionally modify the simulations dataframe to create new grouping variables
+  if(!is.null(make_groups)) simulations_df <- 
+      eval(parse(text = paste("fabricate(simulations_df, ", make_groups, ")")))
```
nfultz: Instead of base NSE using eval/parse - use rlang quosures and splice in make_groups.

Something like

    eval_tidy(quo(fabricate(simulations_df, !!make_groups)))

nfultz (commented):
Agree that NSE makes more sense here.

R/diagnose_design.R
```diff
@@ -94,18 +102,31 @@ diagnose_design <- function(...,
...
-
-  group_by_set <- c("design_label", "estimand_label", "estimator_label", "term")
-
-  if (!is.null(add_grouping_variables)) {
-    group_by_set <- c(group_by_set, add_grouping_variables)
+  
+  # figure out what to group by---------------------------------------------
+  # Optionally modify the simulations dataframe to create new grouping variables
+  if (!is.null(make_groups)) {
+    make_groups_names <- names(make_groups)
```
nfultz, line 109: If the intent is to require named args with `vars()` function you can do the checking/cleaning in the vars function instead of diagnose_design

acoppock: Thanks @nfultz -- here the goal is to allow users to both indicate an existing variable as teh grouping varibale (so would be unnamed) and to create a new grouping variable (so is named)

acoppock:
merged into new-declarations manually

PR #471, opened 2021-06-04
tidy diagnosis

graemeblair:
- aim is to "tidy" a diagnosis and provide a data frame where a row represents an estimate of a diagnosand along with its standard error and a confidence interval for easy graphing and summarizing


R/diagnosis_helper_functions.R
```diff
@@ -208,3 +208,83 @@ clean_bootstrap_df <- function(diagnosis, digits, diagnosand_columns,
...
+    
+    group_by_list <- diagnosand_replicates[, group_by_set, drop = FALSE]
+    
+    labels_df <- split(group_by_list, lapply(group_by_list, addNA), drop = TRUE)
+    labels_df <- lapply(labels_df, head, n = 1)
+    
+    use_vars <- setdiff(names(diagnosand_replicates), c(group_by_set, "bootstrap_id"))
+    diagnosands_summary_df <- split(diagnosand_replicates[use_vars], lapply(group_by_list, addNA), drop = TRUE)
+    
+    diagnosands_mean_list <- lapply(diagnosands_summary_df, function(data) lapply(data[diagnosand_columns], mean))
```
nfultz: I think either a helper function or for-loop would be appropriate here, easier to follow / modify later compared to 4x nested lapplys + unlists.

R/diagnosis_helper_functions.R
```diff
@@ -208,3 +208,83 @@ clean_bootstrap_df <- function(diagnosis, digits, diagnosand_columns,
...
+        conf.high = diagnosands_conf_high_df, 
+        row.names = NULL)
+  }
+  
+  diagnosand_df
+  
+}
+
+
+quantile_NAsafe <- function(x, ...) if(any(is.na(x))) NA else quantile(x, ...)
```
nfultz: `anyNA()`

R/diagnosis_helper_functions.R
```diff
@@ -208,3 +208,83 @@ clean_bootstrap_df <- function(diagnosis, digits, diagnosand_columns,
...
+    diagnosands_se_list <- lapply(diagnosands_summary_df, function(data) lapply(data[diagnosand_columns], sd))
+    diagnosands_conf_low_list <- lapply(diagnosands_summary_df, function(data) lapply(data[diagnosand_columns], quantile_NAsafe, 0.025))
+    diagnosands_conf_high_list <- lapply(diagnosands_summary_df, function(data) lapply(data[diagnosand_columns], quantile_NAsafe, 0.975))
+    
+    diagnosands_mean <- unlist(diagnosands_mean_list)
+    diagnosands_se <- unlist(diagnosands_se_list)
+    diagnosands_conf_low_df <- unlist(diagnosands_conf_low_list)
+    diagnosands_conf_high_df <- unlist(diagnosands_conf_high_list)
+    
+    names <- strsplit(names(diagnosands_mean), split = ".", fixed = TRUE)
```
nfultz: maybe call it labels instead of names, names is also a function you are using.

R/diagnosis_helper_functions.R
graemeblair: That is fair. :) Will fix. Thanks for the read.

PR #481, opened 2022-01-25
CRAN prep 0.30.0

### LINKED ISSUE #409: diagnose_design should automatically record timing info
It would be very informative and mildly useful for me if we also recorded how long the diagnoses took when looking at a bunch of output.

graemeblair:
1. Add `tidy.diagnosis` function.

2. Proposal to make a change to the default diagnosands to (a) reflect current practice and (b) put them in a more logical order.

```
default_diagnosands <- 
  declare_diagnosands(
   mean_estimand = mean(estimand),
   mean_estimate = mean(estimate),
   bias = mean(estimate - estimand),
   sd_estimate = sqrt(pop.var(estimate)),
   rmse = sqrt(mean((estimate - estimand) ^ 2)),
   power = mean(p.value <= alpha),
   coverage = mean(estimand <= conf.high & estimand >= conf.low)
  )
```

(Additional diagnosands are also now emphasized in the help file.)

3. Add diagnosis duration (closes https://github.com/DeclareDesign/DeclareDesign/issues/409)


nfultz:
very minor, but you may want to switch column position of estimand /
estimate, so that visually, col 1 - col 2 = col 3.

PR #487, opened 2022-06-26
add parameter names to sort_by_list

macartan:
parameter names were getting lost with `reshape_diagnosis`  when select invoked see below

note also that the final test in   test-reshape-diagnosis is not reliable as it may produce a vector of length 2 or 3 depending on simulations (`expect_equal(diagnosis$diagnosands_df$significant, structure(c(1L, NA), .Label = "FALSE", class = "factor"))`); this has not be changed here however

e.g.

```
N = 2
design <-  
  declare_model(N = N, Y  = 1:N) +
  declare_estimator(Y ~ 1)

diagnosis <-
  redesign(design, N = 2:3) |> diagnose_design(sims = 3)

reshape_diagnosis(diagnosis, select = "Power")
```

was previously dropping the `N` column


graemeblair:
looks good, will review and try to fix the unreliability you mention

PR #488, opened 2022-06-26
avoiding errors when no estimator provided

macartan:
previously errors for designs that don't have estimators, like below. But there can be interest in understanding properties of estimands without estimators:

```
  design <- 
    declare_model(N = 2) + 
    declare_inquiry(Q = 1) 
  
    design |>
    diagnose_design(sims = 2) |>
    reshape_diagnosis()
```


graemeblair:
LGTM thanks @macartan

PR #491, opened 2023-10-12
Add option to turn off runtime calculation in `diagnose_design`

svraka:
Automatic runtime calculation is a nice addition but it can cause problems, if e.g. you use [targets](https://docs.ropensci.org/targets/) to manage your project pipeline because duration changes between runs and breaks reproducibility.

This PR adds an option to turn off runtime calculation in diagnoses but keeps the old behaviour as default. 

I didn't change anything in printing diagnosis summaries, now with `duration = FALSE` it simply returns `Diagnosis completed in NA seconds.`


nfultz:
Overall seems small/harmless, LGTM. 

On the other hand, I can't see this as something people want to toggle often, maybe using `options()` is a better path than a function parameter?

svraka:
> On the other hand, I can't see this as something people want to toggle often, maybe using `options()` is a better path than a function parameter?

I agree but currently the package doesn't use `options()`, thus I didn't want to introduce any.

Where should it be documented? Would it be sufficiently discoverable in the `diagnose_design()` docstring?

Edit: sorry for closing, clicked the wrong button.

PR #503, opened 2025-09-04
Parallel warnings

macartan:
Modifications get rid of the various serialize warnings by changing the way that objects are stored in designs

Objects now get saved in the relevant quosures and summary of location of where they are saved is added to summary. 
Only objects that are needed are saved. The list of saved objects is reported in model summaries.

```
N <- 5
m <- 4
q <- .5
W <- 6

design <- 
  declare_model(N = N, X = rnorm(N, m)) +
  declare_measurement(M = m) + 
  declare_inquiry(Q = q*N)

design


Research design declaration summary

Step 1 (model): declare_model(N = N, X = rnorm(N, m)) --------------------------

Step 2 (measurement): declare_measurement(M = m) -------------------------------

Step 3 (inquiry): declare_inquiry(Q = q * N) -----------------------------------

Run of the design:

 inquiry estimand
       Q      2.5

Parameters saved in design environments:
# A tibble: 3 × 3
  name  value_str steps
  <chr> <chr>     <chr>
1 N     5         1    
2 m     4         1, 2 
3 q     0.5       3    
draw_estimand(design)

  inquiry estimand
1       Q      2.5
```

## New approach to redesign

Redesigning replaces objects in the relevant quosures:

```
design <- design |> redesign(q = 1)
design


Research design declaration summary

Step 1 (model): declare_model(N = N, X = rnorm(N, m)) --------------------------

Step 2 (measurement): declare_measurement(M = m) -------------------------------

Step 3 (inquiry): declare_inquiry(Q = q * N) -----------------------------------

Run of the design:

 inquiry estimand
       Q        5

Parameters saved in design environments:
# A tibble: 3 × 3
  name  value_str steps
  <chr> <chr>     <chr>
1 N     5         1    
2 m     4         1, 2 
3 q     1         3    
```

Erroneous redesigning signaled

```
design <- design |> redesign(q = 1, W = 2)

You requested a change to W but W is not found in the design
```


nfultz:
This is a fairly large (in LoC) change, but seems fine to the extent that I can follow it. One thing I'd mention, if some public functions are deprecated, maybe add `.Deprecated()` warning and update the roxygen as in https://www.r-bloggers.com/2017/10/r-deprecate-functions-with-roxygen2/ .

I'd also check if the three extra dependencies are providing enough value for the weight. They substantially increase the risk the package gets Ripley'd. Or, if you all do want to be a dplyr package, there's a lot of older code that could be updated to use it, that would be a good project for a grad student.

macartan:
Thanks Neal; the dependencies may not be needed
I’ll look into that

macartan:
* dependencies removed
* no public functions newly deprecated
* everything is backwards compatible
* this pull requests includes elements from two earlier ones that are now withdrawn
* note tests involving DesignLibrary will fail unless you also update that; these two upgrades depend on each other; if you compile both packages you can then run tests on both packages

PR #506, opened 2026-02-18
Improve error message when non-design args passed to diagnose_design

setgree:
diagnose_design() accepts ... which gets forwarded to simulate_design(). Named arguments like alpha = 0.01 that aren't in the function signature silently end up in ... and trigger a cryptic error from check_design_class(). This change identifies which arguments caused the problem and, for the common case of alpha, suggests the correct approach via declare_diagnosands().

### DIFF
--- a/R/diagnose_design.R
+++ b/R/diagnose_design.R
@@ -351,10 +351,22 @@ merge_param_df <- function(df, parameters_df) {
 }
 
 check_design_class <- function(designs) {
-  if (!all(sapply(designs, function(x) {
+  is_valid <- vapply(designs, function(x) {
     inherits(x, "design") || (inherits(x, "function") && is.null(formals(x)))
-  }))) {
-    stop("Please only send design objects or functions with no arguments.")
+  }, logical(1))
+  if (!all(is_valid)) {
+    bad_names <- names(designs)[!is_valid]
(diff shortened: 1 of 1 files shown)


setgree:
wait I meant to only open this on my local branch and then evaluate, sorry!! anyway this was entirely claude code generated so take that for what it's worth

macartan:
hi seth:

looks to me like it is behaving correctly when called correctly; but help could probably be improved and also more informative messages?
```
design <-
  declare_model(
    N = 100,
    Z = complete_ra(N),
    Y = 0.5 * Z + rnorm(N)
  ) +
  declare_estimator(Y ~ Z, inquiry = "ATE")

simulations <- simulate_design(design, sims = 10)

# (below is on simulations_df to acoid redrawing)

# Correct behavior as diagnosands redefined
diagnose_design(simulations, diagnosands = declare_diagnosands( alpha = .01))
diagnose_design(simulations, diagnosands = declare_diagnosands( alpha = .05))
diagnose_design(simulations, diagnosands = select_diagnosands("power", alpha = .01))
diagnose_design(simulations, diagnosands = select_diagnosands("power", alpha = .05))

# Nothing happening here: so not warning about irrelevant argument
diagnose_design(simulations, alpha = .01)

# Error here as only designs expected for unnamed args
diagnose_design(design, alpha = .01)
```

PR #507, opened 2026-04-01
Don't capture expressions of forced arguments with `enquo()`

lionel-:
`enquo()` capture values rather than expressions for forced arguments.

`+` dispatch does force arguments, and capturing the expression only used to work by luck, due to the peculiar way R forces arguments during dispatch. The next version of rlang (to be released soon) uses the new C API of R to capture expressions, which causes DeclareDesign to fail checks because of this.

### DIFF
--- a/R/construct_design.R
+++ b/R/construct_design.R
@@ -1,4 +1,4 @@
-#' Declare a design 
+#' Declare a design
 #'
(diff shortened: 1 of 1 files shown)


nfultz:
@graemeblair We probably should get this merged + uploaded to CRAN ahead of the next major R release, it looks pretty reasonable.

> The release of 4.6.0 ("Happy Hop") is scheduled for Friday 2026-04-24. Prerelease source tarballs (and binaries for some platforms) are being made available during the 4 weeks until the release.

acoppock:
Thanks for this fix, @lionel- -- merged into the CRAN patch at b250889. This
  will ship with v1.1.1.

PR #508, opened 2026-04-06
cran patch 1.1.1

acoppock:
@graemeblair pls review


R/select_diagnosands.R
```diff
@@ -0,0 +1,101 @@
...
+#' diagnose_design(design,
+#'                 diagnosands = select_diagnosands("power", alpha = .00001),
+#'                 sims = 100)
+
+
+
+
+select_diagnosands <- function(..., alpha = 0.05, subset = NULL, na.rm = FALSE) {
+  
+  x <- list(...) |> unlist()
```
nfultz, line 67: Although there is generally too much NSE in this package, I feel like it might be nice to document for this function that is /isn't/. eg

```
neal_diag <- c("power", "type_s_rate")
select_diagnosands(neal_diag, alpha=.01)
```

which ought to work bc of the defensive unlist.

R/expand_design.R
```diff
@@ -70,7 +71,10 @@ expand_design <- function(designer, ..., expand = TRUE, prefix = "design") {
   ix <- lapply(args, seq_along)
   ix <- if(expand) expand.grid(ix) else data.frame(ix)
   
-  designs <- lapply(transp(args, ix), do.call, what = designer)
+  # designs <- lapply(transp(args, ix), do.call, what = designer)
+  jobs <- transp(args, ix)
+  
+  designs <- lapply(jobs, function(dots) {do.call(designer, dots)})
```
nfultz, line 77: No need for the anonymous fn here.

